PR #15, opened 2023-05-23
doc: update instructions to use the helm repo rather than local charts

obs-gh-jeremyjackins:


wyatt-observeinc (approved):
Update makes sense.

PR #19, opened 2023-06-19
feat: use observe endpoint to auto-configure scheme/host/port/customerID 

obs-gh-jeremyjackins:
Rather than generating collector URLs using a provided scheme, customer, host, and port, also allow the entire ~~address~~ endpoint (minus the URL path, which changes per agent) to be provided as a single value. If this value is provided, it overrides the others. This change also adds enforcement that if the full address is not provided, then the customer ID must be provided or the template generation will fail with a descriptive error.

This is required for (still WIP) acceptance tests, which configure the agents to use a local collector whose hostname does not follow the usual pattern (customerID.collect.observeinc.com).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Observe Helm Charts
 
-This repository contains Helm charts for installing the telemetry agents required for Observe Kubernetes apps.
+This repository contains Helm charts for installing the telemetry agents required for Observe apps on Kubernetes.
 
 Contents:
 * stack: Installs several agents required for the Kubernetes Observe app
@@ -19,19 +19,23 @@ helm repo update
 ```
 
 ## Required Values
-You must set `global.observe.customer`. To have Helm create a Kubernetes secret containing your
-datastream token, you must also set `observe.token.value`. Otherwise, you must set `observe.token.create`
-to `false`, and manually create the secrets (see [Managing Secrets Manually](#managing-secrets-manually)).
+You must set `global.observe.collectionEndpoint`, which is provided when configuring a connection in
+Observe. To have Helm create a Kubernetes secret containing your datastream token, you must also set
+`observe.token.value`. Otherwise, you must set `observe.token.create` to `false`, and manually create
+the secrets (see [Managing Secrets Manually](#managing-secrets-manually)).
 
 These values can be set in a custom values file:
 
 ```yaml
 global:
   observe:
-    customer: "123456789012"
+    # A unique URL associated with your customer ID, provided in the new connection installation instructions.
+    collectionEndpoint: "https://123456789012.collect.observeinc.com"
 
 observe:
   token:
+    # The Datastream token. This will typically be unique to each chart being installed, or each release of
+    # of a chart.
     value: <datastream token>
 ```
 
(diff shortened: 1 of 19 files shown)


charts/internal/events/templates/_address.tpl
```diff
@@ -0,0 +1,12 @@
+{{- define "observe.collectorAddress" -}}
```
jta, line 1: We need an env var name we use everywhere. I was leaning towards `OBSERVE_COLLECT_URL`, which would imply we should call this `observe.collectUrl` here, but open to suggestions.

obs-gh-jeremyjackins: I chose "address" rather than URL because this does not include the path. Since address is a more general term I think it's correct here, whereas URL would imply that they should give the full API path. In other words a URL identifies a resource and this address is just a component of that. Let me know what you think.

jta: Yea, this is always going to be a bit prone to confusion.  We have two components:
- the base URL for the API (set by the user)
- the path of the endpoint you intend to use (configured by the developer based on the data, `/v1/http`, `/v1/prometheus`, etc)

There's no reason to constrain the collector URL to expect just the domain name ("address"), and in practice your change doesn't really assume that (I think)? It's useful to be able to point at non-standard address for the purposes of local testing (e.g. "use this ngrok tunnel")

bendrucker: I would avoid "address" except when referring to things explicitly described as addresses: IPs and MACs. Address is much less commonly used for hostnames. Once DNS and load balancers get involved the metaphor of an address as a unique locator for a specific machine is stretched thin. This is still a URL even if the path is extended in individual contexts. If you really want to call it a "base URL" you can but as long as the long form documentation is clear about this I don't think `base` is needed in variable names.

obs-gh-jeremyjackins: Hmm I see address is a more generic term, such as "email address", or "TCP listen address" e.g. `0.0.0.0:8080` when configuring an HTTP server, etc. But OK, I see your point that really this is a URL with the path truncated.

obs-gh-jeremyjackins: > observe.collectUrl

@jta I think the ship may have already sailed for `collect` vs `collector` in these values, as we unfortunately have customers already doing `observe.collectorHost` `observe.collectorPort` etc for these helm charts, so we'd break them if we changed to `collectHost` `collectPort` etc. And I think it would be an annoyance to people to have them inconsistent like `collectorPort` `collectorHost` `collectURL`.

obs-gh-jeremyjackins: On the other hand, if we do decide that it's worth a break, I guess the sooner the better.

bendrucker: Both of those are valid. URLs are addresses:

https://en.wikipedia.org/wiki/Address_space#Examples

Listen address is a good example of where address should be used, as there's no specific name for the `hostname:port` address format.

To be more practical, if you expect a protocol, I'd always call it a URL.

obs-gh-jeremyjackins: @bendrucker counterpoint there is that socket addresses are defined as protocol + IP + port

jta (approved):

PR #24, opened 2023-07-11
fix(all): prevent comment from being incorrectly rendered in endpoint

### LINKED ISSUE #23: observe.collectionEndpoint not generating a valid configuration for  grafana/agent
/charts/internal/endpoint/templates/_address.tpl

generates a config file with:

```
remote_write:
        - url: # re-constructing the URL this way eliminates any path that was included in the valuehttps://999999999999.collect.observeinc.com/v1/prometheus
```

Which results in a metrics pod that has the error:
error loading config file /etc/agent/config.yaml: url for remote_write is empty

obs-gh-jeremyjackins:
This type of comment is not safe to use in the context of these template functions. It can be included in generated endpoints and cause invalid config.

Fixes OB-20629
Fixes #23

### DIFF
--- a/charts/internal/endpoint/Chart.yaml
+++ b/charts/internal/endpoint/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: endpoint
 description: Observe collection endpoint utility functions
 type: library
-version: 0.1.2
+version: 0.1.3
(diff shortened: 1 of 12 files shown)


bendrucker (approved):

bendrucker:
While this is ok in helpers you should generally use template comments instead which will be safe inside the templates themselves. 

This is also how Helm helpers are supposed to be documented and some editors may offer integration to display the comment a la GoDocs. 

https://helm.sh/docs/chart_best_practices/templates/#comments-yaml-comments-vs-template-comments

PR #26, opened 2023-07-17
Add `host_filter: "true"` to default metrics values. 

nlederman:
Chart renders with an empty value without this.

Further in the values, the configmap passed to grafana references a `Values.prom_config.host_filter` which is never initialized.

In our environment this was resulting in memory spikes for the grafana agent that cause an OOM crash loop. Adding this value in as a default appears to stop this, and brings things in line with the previously-used kustomize definition.

### DIFF
--- a/charts/internal/metrics/values.yaml
+++ b/charts/internal/metrics/values.yaml
@@ -17,7 +17,7 @@ grafana-agent:
       scrape_timeout: 10s
       scrape_body_size_limit: 50MB
       scrape_cadvisor_action: keep
-      host_filter: "true"
+      host_filter: "false"
       # the following metrics are exported with 0 values in default cadvisor installs
       # see
       # - https://github.com/kubernetes/kubernetes/issues/60279
--- a/charts/internal/metrics/values.yaml
+++ b/charts/internal/metrics/values.yaml
@@ -17,6 +17,7 @@ grafana-agent:
       scrape_timeout: 10s
       scrape_body_size_limit: 50MB
       scrape_cadvisor_action: keep
+      host_filter: "true"
       # the following metrics are exported with 0 values in default cadvisor installs
       # see
       # - https://github.com/kubernetes/kubernetes/issues/60279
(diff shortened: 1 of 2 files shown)


obs-gh-jeremyjackins:
Thanks for catching this. Looking at our kustomize manifests, as well as the effect of `host_filter`, it looks like this should be set when we run grafana-agent in daemonset mode, e.g. in the examples/stack/values-xl.yaml config. Can you confirm whether you are running grafana-agent in daemonset mode?

charts/internal/metrics/values.yaml
```diff
@@ -17,6 +17,7 @@ grafana-agent:
       scrape_timeout: 10s
       scrape_body_size_limit: 50MB
       scrape_cadvisor_action: keep
+      host_filter: "true"
```
obs-gh-jeremyjackins, line 20: I believe we should set this to "false" by default, and add a line in `examples/stack/values-xl.yaml` which overrides the value to "true", as that example config uses daemonset mode.

I believe that setting this to true in deployment mode (the default) would mean that any nodes not running a grafana-agent pod will not have metrics scraped.

nlederman: Ah, okay, I knew I shoulda figured out the root of the issue. I was looking at this against how we had things previously set with Kustomize and assumed it should be defaulted. I was thinking last night after I left the office that I shoulda figured out the actual usage of that before just assuming.

Yes, we are running this in Daemonset mode. So that would make sense then why we would need it but in a deployment you wouldn't want it.

examples/stack/values-xl.yaml
```diff
@@ -39,6 +39,8 @@ metrics:
                   - {key: kubernetes.io/os, operator: NotIn, values: [windows]}
 
     agent:
+      prom_config:
+        host_filter: "true"
```
obs-gh-jeremyjackins, line 43: The template is looking for `.Values.prom_config.host_filter`, but this change would add `.Values.agent.prom_config.host_filter`. These two lines need to be moved out of the agent block -- you can put it just under `grafana-agent` before the `controller` block.

obs-gh-jeremyjackins:
Thank you for the fix!

PR #27, opened 2023-07-18
chore: making event chart more flexible

muratso:
This PR adds:
* custom environments variables via extraEnv
* custom toleration for the events deployment
* custom imagePullPolicy for the events deployment images
* custom command and arguments if needed. If not specified, the default ones from the previous version will be used
* possibility to not use the `envFrom: - secretRef` if a user wants to specify this directly via extraEnvs or other means
* possibility to use a custom `secretRef` name if needed/wanted
* also added a few default values in the values.yaml allowing users to properly run `helm template` and `helm lint` with the chart out-of-the-box.

### DIFF
--- a/charts/internal/events/templates/deployment.yaml
+++ b/charts/internal/events/templates/deployment.yaml
@@ -75,9 +75,11 @@ spec:
             {{- range .Values.extraEnv }}
             - {{- toYaml . | nindent 14 }}
             {{- end }}
+          {{- if .Values.useCredentialsSecret.enabled }}
           envFrom:
             - secretRef:
-                name: credentials
+                name: {{ .Values.useCredentialsSecret.secretName }}
+          {{- end}}
           readinessProbe:
             httpGet:
               path: /healthz
--- a/charts/internal/events/templates/deployment.yaml
+++ b/charts/internal/events/templates/deployment.yaml
@@ -72,6 +72,9 @@ spec:
                 configMapKeyRef:
                   name: cluster-info
                   key: id
+            {{- range .Values.extraEnv }}
+            - {{- toYaml . | nindent 14 }}
+            {{- end }}
           envFrom:
             - secretRef:
                 name: credentials
--- a/charts/internal/events/templates/deployment.yaml
+++ b/charts/internal/events/templates/deployment.yaml
@@ -17,6 +17,10 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+      {{- with .Values.tolerations }}
+      tolerations:
+      {{- toYaml . | nindent 6 }}
+      {{- end }}
       affinity:
         nodeAffinity:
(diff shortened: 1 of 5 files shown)


charts/internal/events/values.yaml
```diff
@@ -1,13 +1,20 @@
+global:
+  observe:
+    customer: "123456789012"
```
obs-gh-jeremyjackins, line 3: Can you please remove this global block -- the customer ID is used to generate URLs, which is now done with a template include. global.observe.customer is still a supported way to configure the chart, for backwards-compatibility, but it's now deprecated in favor of `collectionEndpoint`.

charts/internal/events/templates/deployment.yaml
```diff
@@ -36,6 +41,17 @@ spec:
       containers:
         - name: kube-state-events
           image: {{ .Values.image.kube_state_events.repository }}:{{ default .Chart.AppVersion .Values.image.kube_state_events.tag }}
+          imagePullPolicy: {{ .Values.image.kube_state_events.imagePullPolicy | default "IfNotPresent" }}
+          {{- if .Values.image.kube_state_events.command }}
+          command: {{- range .Values.image.kube_state_events.command }}
```
obs-gh-jeremyjackins, line 46: Do you have a use case in mind for overriding the command and args?

obs-gh-jeremyjackins:
Thanks for the submission! Added some initial feedback, and will look more closely as time allows.

charts/internal/events/values.yaml
muratso: Hey @obs-gh-jeremyjackins The only reason I set it is that it is "required" in order to run a `helm template` or `helm lint`. If I don't have a default value, even a placeholder like I did, these commands can't be executed properly. If you prefer I can set a default placeholder for the `collectionEndpoint` instead of this.

charts/internal/events/templates/deployment.yaml
muratso: Yep. Our use case. We use some sort of "wrapper" to inject sensitive information/variables into pods. In order to have this wrapper executed, we need to run a different entrypoint. This change won't affect the current behavior but will allow us (and other possible customers) to make these customizations without having to fork and maintaining their own version of it (like we're doing right now)

muratso:
@obs-gh-jeremyjackins just circling back here. Let me know if I need to change anything before I rebase my branch and ask for another review. Thanks!

obs-gh-colinhutchinson:
@muratso sorry for the delay.

I picked up some of your changes in a PR that avoids the conflicts.

muratso:
> @muratso sorry for the delay.
> 
> I picked up some of your changes in a PR that avoids the conflicts.

No worries. We can probably close this PR and I'll check what has been done. If needed, I'll open a new one, or keep doing like I did (which is pretty much having our own fork).

PR #28, opened 2023-07-19
feat: linting and end-to-end collection tests

obs-gh-jeremyjackins:
This PR is large for a few reasons:
- Making the linter happy required some diff-heavy changes; for example it did not like some charts being organized into a directory called `internal`, so those were moved to the top level `charts` directory.
- Some real bugs were caught and fixed
- It's actually a fairly large change, introducing two new programs: the test collector, and the test client. Luckily neither are complex.

It also introduces a new chart which is currently only used for tests, called `proxy`. This chart lets us run a local collector which can act as a proxy for all events emitted by the various agents. The only current (perhaps the only ever) use case is for testing, so that we can examine the data coming from the agents and ensure they are behaving as expected. It does not actually proxy the data to anywhere, in this test use case. It's running the new test collector mentioned above. Other than for testing, this subchart is disabled by default for `stack` and `traces`.


bendrucker:
Reviewing this today!

test/cmd/go.mod
```diff
@@ -0,0 +1,34 @@
...
+	go.opentelemetry.io/otel/metric v1.16.0 // indirect
+	go.opentelemetry.io/otel/trace v1.16.0 // indirect
+	go.opentelemetry.io/proto/otlp v0.19.0 // indirect
+	golang.org/x/net v0.8.0 // indirect
+	golang.org/x/sys v0.8.0 // indirect
+	golang.org/x/text v0.8.0 // indirect
+	google.golang.org/genproto v0.0.0-20230306155012-7f2fa6fef1f4 // indirect
+	google.golang.org/grpc v1.55.0 // indirect
+	google.golang.org/protobuf v1.30.0 // indirect
+)
```
obs-gh-colinhutchinson, line 34: (optional) setup dependabot

.github/workflows/test.yml
```diff
@@ -0,0 +1,51 @@
+name: Test Charts
```
bendrucker, line 1: Minimize name repetition to maximize the use of space in status check names. I'd say you can drop "charts" from the names (workflow and job), since that's implied by the repository name. `test/test` (same workflow/job name) is unavoidable and I wouldn't worry about that.

.github/workflows/test.yml
bendrucker, line 1: Enable Dependabot for Actions, particularly because there's several exact action versions in here, and you want to get bug fixes.

bendrucker (approved):
Nice work here!

obs-gh-colinhutchinson:
Feel free to punt this to a separate PR.

https://github.com/hckops/actions#helm-dependencies-action

If I'm not mistaken everything in `test/cmd/` kind of accomplishes [environment checking script](https://www.notion.so/Integration-Predictability-ca17134656e944e6874f467fe3b73419?d=357205a53f3e410489a28965326f53d0&pvs=4#001ac01dee9642b38d90ee8097ab48b3)? Is this a tool future us can productionize / generalize to other situations?

obs-gh-jeremyjackins:
> If I'm not mistaken everything in `test/cmd/` kind of accomplishes [environment checking script](https://www.notion.so/Integration-Predictability-ca17134656e944e6874f467fe3b73419?d=357205a53f3e410489a28965326f53d0&pvs=4#001ac01dee9642b38d90ee8097ab48b3)? Is this a tool future us can productionize / generalize to other situations?

Thanks for pointing me at this document, left a comment there. Yes, I definitely think we can use this in other situations. I already have plans to move the code to the monorepo (that's where the collection decoder library lives, and I want to be able to query the decoded contents), which would make it easier to use it in other places as well. But Helm is a good first use case to iron out the kinks.

PR #32, opened 2023-08-14
fix: add test hook annotations to test ConfigMaps

bendrucker:
* Adds test hook annotations to ConfigMaps used in tests to prevent installation during a normal `helm install`
* Adds test coverage by also installing charts together, but not waiting. This will detect any naming conflict which might arise between the charts, this case just happened to be test resources.

### DIFF
--- a/charts/stack/templates/tests/test-stack.yaml
+++ b/charts/stack/templates/tests/test-stack.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: queries
+  annotations:
+    "helm.sh/hook": test
 data:
   # Queries have the format <JMESPath query> -> <matching regexp>.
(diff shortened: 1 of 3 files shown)


bendrucker:
Just adding a hook annotation isn't sufficient. That essentially makes the hook a "test" of its own which gets rolled back after completion. Even with a weight of `-1`, that just means the ConfigMap gets created (and then deleted) first. Setting a [deletion policy](https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies) could preserve it but the whole point of `helm test` is that it shouldn't have leftovers.

The addition to `test.sh` can be kept but the config will need to be inlined somehow, e.g.,:

* Passed via env vars instead of files
* Generated by an init container into a shared volume

obs-gh-jeremyjackins:
Sent https://github.com/observeinc/helm-charts/pull/36 which removes the configmaps.

PR #36, opened 2023-08-22
fix: remove configmaps from tests

obs-gh-jeremyjackins:
Each kubernetes manifest is treated as a standalone test by helm, so creating configmaps to configure test pods does not work as expected -- configmaps either persist, which is a bug, or they are cleaned up before the job relying on them can be run. Avoid them entirely for the configuration of test pods.

Closes #32

### DIFF
--- a/charts/traces/templates/tests/test-traces.yaml
+++ b/charts/traces/templates/tests/test-traces.yaml
@@ -1,24 +1,4 @@
 apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: queries
-data:
-  # Queries have the format <JMESPath query> -> <matching regexp>.
-  # This syntax is based on the syntax used in the JMESPath docs, which describe queries as:
-  # search(query, input) -> result
-  queries: |
-    Path -> ^/v1/otel/v1/traces$
-
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: cluster-info
-data:
-  id: "fake-cluster-id"
-
----
-apiVersion: v1
 kind: Pod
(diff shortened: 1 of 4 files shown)


charts/traces/templates/tests/test-traces.yaml
```diff
@@ -1,24 +1,4 @@
...
-  queries: |
-    Path -> ^/v1/otel/v1/traces$
-
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: cluster-info
-data:
-  id: "fake-cluster-id"
```
obs-gh-jeremyjackins, line 18: known regression: this (traces) test will now fail if you run it in isolation, as it depends on the `events` pod (which manages the creation of this configmap) of the `stack` test having run first. Luckily it is normally not run in isolation, as both `make test` and the github test automation run the `stack` and then `traces` test in order.

There is another open issue to have `traces` run its own init container to set up cluster-info, which will fix this regression.

obs-gh-jeremyjackins:
Closes #32.

bendrucker (approved):
Very nice! I like the arg/flag approach over files anyway regardless of the underlying need. Easier to share reproducible examples, doesn't seem like there's enough config to require files.

FYI I edited your PR's body to move "Closes" there. You'll see an underline and tooltip when it's recognized. It works in merged PR bodies and commit messages in the default branch but not PR comments, since that would be too subtle and introduce authorization complexity.

obs-gh-jeremyjackins:
Ah, thanks for catching the Closes issue.

PR #44, opened 2023-09-18
chore(ci): validate Chart.yaml is updated when necessary

obs-gh-colinhutchinson:
- add a bash script / Make task to update Chart.yaml ( [example](https://github.com/hutchic-observe-meta/helm-charts/pull/1/commits/e35599127102f5a1926837c4d24c303af038196c) )
- add a github actions to spot check that Chart.yaml is updated appropriately ( [example](https://github.com/hutchic-observe-meta/helm-charts/actions/runs/6224892802/job/16894112642) )

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -32,16 +32,4 @@ jobs:
 
       - name: Check if Chart.yaml versions are bumped if relevant
         run: |
-          CHARTS="endpoint events logs metrics proxy stack traces"
-          for chart in $CHARTS; do
-            # Check if files within the chart directory have changed
-            if git diff --name-only origin/main...HEAD | grep -E "^charts/$chart/"; then
-              echo "Changes detected in charts/$chart/"
-
-              # Check if the Chart.yaml for this chart also has changes
-              if ! git diff --name-only origin/main...HEAD | grep -E "^charts/$chart/Chart.yaml"; then
-                echo "Error: Detected changes in charts/$chart/ but no changes in Chart.yaml."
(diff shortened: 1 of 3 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -23,3 +23,25 @@ jobs:
...
           token: ${{ secrets.GITHUB_TOKEN }}
+
+  check-chart-version-bump:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4.0.0
+
+      - name: Check if Chart.yaml versions are bumped if relevant
+        run: |
```
jta, line 34: We should keep the action relatively simple and simply invoke `make check` or something of the sort. That would allow users to replicate locally and evolve the logic without touching the workflow.

.github/workflows/pre-commit.yml
```diff
@@ -23,3 +23,25 @@ jobs:
...
+          CHARTS="endpoint events logs metrics proxy stack traces"
```
jta, line 35: I can't imagine there's a reason to not check every single chart, in which case a `find` here looking for Chart.yaml might be more appropriate.

jta (changes requested):

.github/workflows/pre-commit.yml
obs-gh-colinhutchinson: yeah I considered that .... I don't think I had a reason not to do it that way then I just didn't lol

jta (approved):

PR #47, opened 2023-09-21
fix(events): make events liveness probe more forgiving

obs-gh-colinhutchinson:
```
diff /tmp/new.yaml /tmp/old.yaml         
8c8
<     helm.sh/chart: events-0.1.11
---
>     helm.sh/chart: events-0.1.10
418c418
<     helm.sh/chart: events-0.1.11
---
>     helm.sh/chart: events-0.1.10
489c489
<     helm.sh/chart: events-0.1.11
---
>     helm.sh/chart: events-0.1.10
674c674
<     helm.sh/chart: events-0.1.11
---
>     helm.sh/chart: events-0.1.10
742,744c742,743
<             initialDelaySeconds: 30
<             timeoutSeconds: 10
<             failureThreshold: 20
---
>             timeoutSeconds: 5
>             failureThreshold: 10
```

Also improved some of the automation
related: https://github.com/observeinc/manifests/pull/118


wyatt-observeinc (approved):
LGTM

charts/events/templates/deployment.yaml
```diff
@@ -73,7 +73,8 @@ spec:
             httpGet:
               path: /healthz
               port: 5171
-            timeoutSeconds: 5
-            failureThreshold: 10
+            initialDelaySeconds: 30
```
jta, line 76: I suspect this is unnecessary. This will delay the initial probe, which means that pods will never be ready within this interval. The observe-events pod would never be ready within 30 seconds even for small clusters. I think increasing `failureThreshold` and / or `periodSeconds` achieves the intended result without impacting smaller clusters.

charts/events/templates/deployment.yaml
```diff
@@ -73,7 +73,8 @@ spec:
...
+            timeoutSeconds: 10
```
jta, line 77: I don't think I've ever seen a timeout. 10 seconds sounds like an extremely long time to return a reply. Coupled with failure threshold we'd get an extremely long worst-case liveness failure.

jta (approved):

bump_version.sh
```diff
@@ -7,9 +7,28 @@ if [ -z "$CHART_PATH" ]; then
...
 CURRENT_VERSION=$(echo $CURRENT_VERSION_LINE | awk '{print $2}')
+MAIN_VERSION=$(git show origin/main:$CHART_PATH/Chart.yaml | awk '/^version:/ {print $2}')
+
+# If the version has already been bumped compared to main, exit
+if [ "$CURRENT_VERSION" != "$MAIN_VERSION" ]; then
+    echo "Version in $CHART_PATH/Chart.yaml has already been bumped, skipping bump."
+    exit 0
+fi
+
 BUMPED_VERSION=$(echo $CURRENT_VERSION | awk -F. '{$NF = $NF + 1;} 1' OFS=.)
```
obs-gh-robnavarro, line 32: This awk call has to be the weirdest way of doing: `'BEGIN {OFS="."} {$NF=$NF+1; print}'`
that I have seen :-|

PR #53, opened 2023-10-26
feat: event chart toleration and pull policy

obs-gh-colinhutchinson:
grabbed the low hanging fruit that was proposed in https://github.com/observeinc/helm-charts/pull/27


wyatt-observeinc (approved):
Reviewed with team

PR #61, opened 2023-11-02
chore: setup updatecli and a gha job

obs-gh-colinhutchinson:
Generates update PR's
image

Validates PR's have updated all the necessary Chart.yaml version bumps
image

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
 
-  check-chart-version-bump:
+  chart-version-bumps:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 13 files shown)


updatecli/internal/02_proxy.yml
```diff
@@ -0,0 +1,48 @@
...
+      key: "$.version"
+  updateStackChart:
+    name: "Update Stack Chart Dependency"
+    kind: yaml
+    sourceid: currentProxyVersion
+    transformers:
+      - semverinc: "patch"
+    spec:
+      file: "charts/stack/Chart.yaml"
+      key: "$.dependencies[3].version"
```
obs-gh-colinhutchinson, line 39: a bug in the yaml parsing library means we must use array indexes as it's not possible to search based on `:name=xxx`

.github/workflows/pre-commit.yml
```diff
@@ -24,14 +24,37 @@ jobs:
...
 
-      - name: Check if Chart.yaml versions are bumped if relevant
+      - name: Setup updatecli
+        uses: updatecli/updatecli-action@v2
+
+      - name: Bump any charts that require it
         run: |
-          make validate-chart-version
+          make updatecli/internal
+          git diff > diff.txt
```
jta, line 41: too much logic, why can't this be folded into the makefile?

obs-gh-colinhutchinson: Ideally the github action would do inline code comments that the author could accept to update their PR seamlessly.

In the abasence of being able to do that I thought a comment on the PR was the next best ...

jta (approved):

PR #69, opened 2023-11-08
fix(traces): extract k8s.cluster.uid through k8sattributes

obs-gh-abhishekrao:
k8sattributes processor now natively supports extracting `k8s.cluster.uid`, so we don't need to handle this through the resource processor anymore.

Kustomize PR: https://github.com/observeinc/manifests/pull/131

### DIFF
--- a/charts/traces/values.yaml
+++ b/charts/traces/values.yaml
@@ -9,8 +9,6 @@ proxy:
 opentelemetry-collector:
   fullnameOverride: "observe-traces"
   nameOverride: traces
-  image:
-    tag: "0.62.1"
   mode: "daemonset"  # daemonset or deployment
   service:
     enabled: true
@@ -31,6 +29,7 @@ opentelemetry-collector:
           - ""
         resources:
           - pods
+          - namespaces
         verbs:
           - get
           - list
@@ -73,15 +72,6 @@ opentelemetry-collector:
     jaeger-grpc:
       enabled: false
   extraEnvs:
-    - name: NODE_NAME
-      valueFrom:
-        fieldRef:
-          fieldPath: spec.nodeName
-    - name: OBSERVE_CLUSTER
-      valueFrom:
-        configMapKeyRef:
-          name: cluster-info
-          key: id
     - name: OBSERVE_TOKEN
       valueFrom:
         secretKeyRef:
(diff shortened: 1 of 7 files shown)


charts/traces/values.yaml
```diff
@@ -119,17 +114,13 @@ opentelemetry-collector:
...
-          - key: k8s.cluster.uid
-            value: "${OBSERVE_CLUSTER}"
-            action: insert
       k8sattributes:
         passthrough: false
         extract:
           metadata:
             - k8s.pod.name
             - k8s.namespace.name
+            - k8s.cluster.uid
```
jta, line 123: Do we have the necessary cluster role privileges? How does the processor retrieve the UID?

obs-gh-abhishekrao: Based on what I read [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.88.0/processor/k8sattributesprocessor#role-based-access-control), we would need an additional resource (namespaces) in RBAC. There's a comprehensive example [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.88.0/processor/k8sattributesprocessor/testdata/e2e/collector/configmap.yaml) which also refers to a similar RBAC setup. We don't collect deployment / replicaset attributes so we don't need any more than this.

Additionally, our config was not [filtering](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.88.0/processor/k8sattributesprocessor#as-an-agent) down to the node the daemonset was running in, without which it would try and list every pod across every namespace. I also added that, but let me know if there's a reason for the way it was.

jta: I assume if you need to grab the UID of `kube-system` you will need a cluster role and cluster role binding. I believe we already have that set up as part of the service account because it was required by cluster-info, but we should make sure that is the case and remove the cluster info dependency entirely.

As for filtering, we do not assume otel-collector is run as a daemonset. There are cases where that is not an option (e.g. EKS on Fargate)

obs-gh-abhishekrao: Yes, cluster-info dependency is no longer needed. I did note that we need permissions on `namespaces` in addition to `pods` for the new resource detection to work correctly so I added that.

Also, the Helm defaults to a daemonset, which is why I added the filtering. 
https://github.com/observeinc/helm-charts/blob/7b5df82f49bc2566149a28b48291305c3bc4a321/charts/traces/values.yaml#L13

obs-gh-colinhutchinson (approved):

PR #95, opened 2024-01-24
fix: possible config example

arthur-observe:
Add a new values-minimal.yaml chart

Idea: Maybe we just create a new example and then point to this as a good default but not interfere with any existing?

### DIFF
--- a/examples/stack/values-minimal.yaml
+++ b/examples/stack/values-minimal.yaml
@@ -1,0 +1,45 @@
+global:
+  observe:
+    collectionEndpoint: ""
+
+observe:
+  token:
+    value: ""
+
+events:
+  resources:
+    limits:
+      cpu: 50m
+      memory: 256Mi
+    requests:
+      cpu: 50m
+      memory: 256Mi
+
+logs:
+  fluent-bit:
+    resources:
+      limits:
+        cpu: 100m
+        memory: 128Mi
+      requests:
+        cpu: 100m
+        memory: 128Mi
+    config:
+      mem_buf_limit: 10MB
+
+metrics:
+  grafana-agent:
+    prom_config:
+    scrape_configs:
+      cadvisor_action: drop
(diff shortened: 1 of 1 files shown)


examples/stack/values-minimal.yaml
```diff
@@ -0,0 +1,46 @@
...
+        memory: 128Mi
+    config:
+      mem_buf_limit: 10MB
+
+metrics:
+  grafana-agent:
+    prom_config:
+    scrape_interval: 60s
+    scrape_configs:
+      cadvisor_action: drop
```
NathanNam, line 35: I'm fine with this approach.

Why did we set `cadvisor_action: drop`?

https://observeinc.slack.com/archives/C06EBP24T1V/p1706044118395189

NathanNam: Also the indentation is incorrect.

NathanNam: Let's use this one including dropping `container_fs.*` as well.

```
metrics:
  grafana-agent:
    prom_config:
      scrape_interval: 60s
      scrape_configs:
        pod_action: drop
        cadvisor_metric_drop_regex: container_(network_tcp_usage_total|network_udp_usage_total|tasks_state|cpu_load_average_10s|fs_.*)
        cadvisor_metric_keep_regex: container_(cpu_.*|spec_.*|memory_.*|network_.*|file_descriptors)|machine_(cpu_cores|memory_bytes)
```

NathanNam: BTW, we should make the same change for kustomize as well.

NathanNam: This values-minimal.yaml is based on values-xs.yaml. What do we recommend if customers need to use a large or extra large templates?

arthur-observe:
example only

PR #105, opened 2024-01-31
fix(metrics): drop pod metrics by default

obs-gh-nikhildua:
https://www.notion.so/observeinc/Kubernetes-Follow-up-on-OB-27594-0c3e01f24f554505bde424517870bdbb#1c00281607504f19ada61a3a75565411

Address #2 

Tested and verified in K8s that metrics are dropped 

Screenshot 2024-01-31 at 1 31 31 PM


NathanNam (approved):
LGTM

PR #115, opened 2024-02-07
fix(traces): Makes collection endpoint path configurable

obs-gh-konstantintikhonov:
https://observe.atlassian.net/browse/OB-28334

Introduces a new optional configuration value `global.observe.otelPath` which is set to `/v1/otel` by default. Can be set to `/v2/otel` to send OpenTelemetry observations to v2 version of collection endpoint.


jta (approved):

NathanNam (approved):
LGTM!

PR #131, opened 2024-03-11
doc: added upgrade instructions to use /v2/otel endpoint

obs-gh-konstantintikhonov:
Observe backend supports a new /v2/otel endpoint to receive OpenTelemetry observations and process them more efficiently than the previous /v1 endpoint. 

This PR adds instructions how to configure otel-collector to send observations to the new endpoint

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,19 +68,33 @@ helm install --namespace=observe observe-traces observe/traces \
 helm -n observe get values observe-traces -o yaml > observe-traces-values.yaml
 ```
 
-### Using v2 of collection endpoint
+### Using v2 of OpenTelemetry collection endpoint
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -83,6 +83,20 @@ helm install --namespace=observe observe-traces observe/traces \
...
 ```
 
+When upgrading the `traces` chart:
+
+```
+helm upgrade --namespace=observe observe-traces observe/traces --reuse-values \
+	--set global.observe.otelPath="/v2/otel"
+```
+
+When upgrading the `stack` chart:
```
obs-gh-kaisu, line 93: Does the otelPath parameter have any effect in the stack chart? I thought it's only limited the traces chart.

obs-gh-konstantintikhonov: Yes, users can install traces as a part of stack. In that case, they need to update stack chart

obs-gh-kaisu (approved):

PR #138, opened 2024-05-10
fix(logs): make network properties configurable

jta:


charts/logs/values.yaml
```diff
@@ -65,6 +65,12 @@ fluent-bit:
     grep_match_tag: "nothing"
     grep_exclude: "nomatch ^$"
     inotify_watcher: "true"
+    workers: 2
```
obs-gh-abhishekrao, line 68: This part gives me some doubt - see https://docs.fluentbit.io/manual/v/dev-2.2/pipeline/outputs/stackdriver#occasional-crashing-with-greater-than-1-workers.

obs-gh-abhishekrao: Oh that's stackdriver. Ignore.

obs-gh-abhishekrao (approved):

PR #151, opened 2024-07-11
feat: add observe agent helm charts

obs-gh-alexlew:


charts/agent/charts/daemonset/values.yaml
```diff
@@ -0,0 +1,54 @@
...
+  image:
+    repository: observeinc/observe-agent
+    # Overrides the image tag whose default is the chart appVersion.
+    # The service's name will be applied to the end of this value.
+    tag: "0.7.0"
+    pullPolicy: IfNotPresent
+
+  command:
+    name: "observe-agent"
+    extraArgs: ["start", "--config=/conf/observe-agent.yaml"]
```
arthur-observe, line 17: I am concerned about the idea of never being able to see the config file that is actually in use.  Is it possible to use the code (or some vanilla Go) that assembles the config today to write out a chart to a configmap so it can be used to troubleshoot.  I love generating the config off of the higher level abstraction but I worry a lot about a black box config file.  Why not write something out instead of assembling in memory?

charts/agent/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
```
arthur-observe, line 1: I assume we will have several more revisions of this chart - I think roles will need some switches base on other configurations if we are taking a least privilege approach.

charts/agent/charts/daemonset/values.yaml
arthur-observe: You don't need to resolve this right now but I think we need to have a conversation about how information will be exposed.  The underlying otel collector chart takes config as an argument and produces a config map.  Perhaps we could consider a different approach then the host solution takes.

arthur-observe (approved):

charts/agent/charts/daemonset/values.yaml
obs-gh-alexlew: yup, this is a feature request that's come up a good amount. there are proposals and onging work to add it as an extension to zpages and expose an endpoint that returns it but there's no built in support in otel collector today.

obs-gh-alexlew, line 17: `--otel_config=/path/to/configmap`

PR #161, opened 2024-07-31
fix: remove unused daemonset

obs-gh-alexlew:
Remove unused daemonset subchart from `agent` chart


charts/agent/README.md.gotmpl
```diff
@@ -0,0 +1,23 @@
+{{ template "chart.header" . }}
+{{ template "chart.deprecationWarning" . }}
+
+{{ template "chart.badgesSection" . }}
+
+> [!CAUTION]
```
obs-gh-alexlew, line 6: need to add this template otherwise the caution section gets overwritten

obs-gh-abhishekrao (approved):

PR #171, opened 2024-08-12
chore(ci): use helm-docs instead of pre-commit to auto-update chart documentation

obs-gh-abhishekrao:
Previously, we tried the pre-commit action which does not work, because it only checks for diffs and doesn't behave as it would locally. Since we're mainly solving for helm-docs, we prefer to use that instead.


obs-gh-justindaines (approved):
Using losisin/helm-docs-github-action@v1 seems promising.

PR #175, opened 2024-08-14
fix: ob-35088 update agent version and bump chart version

arthur-observe:
Move agent version to 0.10.0 with has prometheus remote write receiver so we don't need to override image with otel collector


charts/agent/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.4.1
-appVersion: "0.9.0"
+version: 0.4.2
+appVersion: "0.10.0"
```
obs-gh-alexlew, line 6: nit: think this also should be bumped to 0.11.0

charts/agent/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.4.1
-appVersion: "0.9.0"
+version: 0.4.2
+appVersion: "0.11.0"
```
obs-gh-abhishekrao, line 6: Why mix minor and patch version updates?

PR #176, opened 2024-08-15
fix: ob-35088 add entity endpoint for objects

arthur-observe:
1. Add exporter and conditional pipeline section for entity endpoint base on entityToken presence
2. Change ottlp/http to base exporter
3. Add trim for endpoint that handles trailing slash or not

### DIFF
--- a/charts/agent/templates/_deployment-cluster-events-config.tpl
+++ b/charts/agent/templates/_deployment-cluster-events-config.tpl
@@ -5,7 +5,10 @@ extensions:
 
 exporters:
 {{- include "config.exporters.debug" . | nindent 2 }}
-{{- include "config.exporters.otlphttp.observe" . | nindent 2 }}
+{{ if .Values.observe.entityToken }}
+{{- include "config.exporters.otlphttp.observe.entity" . | nindent 2 }}
+{{ end }}
+{{- include "config.exporters.otlphttp.observe.base" . | nindent 2 }}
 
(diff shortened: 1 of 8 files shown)


charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -84,6 +85,23 @@ processors:
 
 {{- include "config.processors.attributes.k8sattributes.podcontroller" . | nindent 2 }}
 
+  # attributes to append to objects
+  attributes/observe_objects:
+    actions:
+      - key: observe_filter
```
xueweiz, line 91: Why are we adding these keys?
They won't be preserved in Observe anyways.

Here is our direct-write entity table design: https://www.notion.so/observeinc/2024-07-k8s-direct-write-entity-table-schema-design-76a762812acf4e028e3f80ded51f0aa3?pvs=4#06d0261ed97740bfa9c1c57454db0e93
You can see we only look at `observe_transform`, we never look at `observe_filter`, let alone store it.

If you want to add debug information, let's just add it to `observe_transform.control`, and prefix it with `debug_`, like this: https://github.com/observeinc/helm-charts/blob/f3ec22a207f05c31fadd4eb885f4a6cea5ee9e81/charts/agent/templates/_deployment-cluster-events-config.tpl#L227

xueweiz (approved):

PR #177, opened 2024-08-15
fix: Overhaul on daemonset and events deployment

xueweiz:
1. fixed pod controller facets
2. enabled hostfs and kubeletstats metric collection properly
3. enabled k8sattributes properly for the kubeletstats metrics

(#2 and #3 was broken because the presets were dropped during the repo migration.

### DIFF
--- a/charts/agent/templates/_config-exporters.tpl
+++ b/charts/agent/templates/_config-exporters.tpl
@@ -1,13 +1,13 @@
 {{- define "config.exporters.otlphttp.observe" -}}
 otlphttp/observe:
-    endpoint: "{{ .Values.observe.collectionEndpoint }}v2/otel"
+    endpoint: "{{ .Values.observe.collectionEndpoint }}/v2/otel"
     headers:
         authorization: "Bearer {{ .Values.observe.token }}"
 {{- end -}}
 
 {{- define "config.exporters.prometheusremotewrite" -}}
 prometheusremotewrite:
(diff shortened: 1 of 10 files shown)


charts/agent/templates/_config-exporters.tpl
```diff
@@ -1,13 +1,13 @@
 {{- define "config.exporters.otlphttp.observe" -}}
 otlphttp/observe:
-    endpoint: "{{ .Values.observe.collectionEndpoint }}v2/otel"
+    endpoint: "{{ .Values.observe.collectionEndpoint }}/v2/otel"
```
arthur-observe, line 3: I have a tweak for this section that makes it so it doesn't matter if endpoint comes in with slash or not.  I can rebase on this to reconcile.

charts/agent/templates/_config-exporters.tpl
```diff
@@ -1,13 +1,13 @@
...
     headers:
         authorization: "Bearer {{ .Values.observe.token }}"
 {{- end -}}
 
 {{- define "config.exporters.prometheusremotewrite" -}}
 prometheusremotewrite:
-    endpoint: "{{ .Values.observe.collectionEndpoint }}v1/prometheus"
+    endpoint: "{{ .Values.observe.collectionEndpoint }}/v1/prometheus"
```
arthur-observe, line 10: ditto

arthur-observe (approved):

PR #189, opened 2024-08-26
feat: cleanup tagging for (mostly) metrics

xueweiz:
1. in k8sattributes, stop adding all labels and annotations as tags. As that is a very pretty big list with unclear value. We should enable them selectively when user needs them.

2. removed k8sattributes/podcontroller, no one is using them

3. dropped resourcedetection for all deployments This is the one that tags a lot of node level information. But, for deployments, it tags about the node that the deployment runs on, instead of the node that runs the things being monitored (e.g. pods). So, the result is very confusing. Let's just drop it. We can add selective tags back when they are useful.

4. change the debug_objectSource tag to debug_source Since it was initially used for entity/object streams only. But now it's being used for logs & metrics data too.

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -14,14 +14,6 @@ batch:
 {{- define "config.processors.attributes.k8sattributes" -}}
 k8sattributes:
   extract:
-    annotations:
-    - from: pod
-      key_regex: (.*)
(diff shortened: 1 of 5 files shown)


obs-gh-enricogiorio (approved):

obs-gh-enricogiorio:
Thanks for the cleanup!

PR #191, opened 2024-08-27
feat: ob35943 create secret for token

arthur-observe:
Creates a template for secret creation on cluster to hold observe token

### DIFF
--- a/charts/agent/templates/_config-exporters.tpl
+++ b/charts/agent/templates/_config-exporters.tpl
@@ -1,22 +1,22 @@
 {{- define "config.exporters.otlphttp.observe.base" -}}
 otlphttp/observe/base:
-    endpoint: "{{ .Values.observe.collectionEndpoint | toString | trimSuffix "/" }}/v2/otel"
+    endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v2/otel"
     headers:
-        authorization: "Bearer {{ .Values.observe.token }}"
+        authorization: "${env:OBSERVE_TOKEN}"
 {{- end -}}
 
 {{- define "config.exporters.otlphttp.observe.entity" -}}
 otlphttp/observe/entity:
-    logs_endpoint: "{{ .Values.observe.collectionEndpoint | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
+    logs_endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
     headers:
-        authorization: "Bearer {{ .Values.observe.entityToken }}"
+        authorization: "Bearer {{ .Values.observe.entityToken.value }}"
 {{- end -}}
 
 {{- define "config.exporters.prometheusremotewrite" -}}
 prometheusremotewrite:
-    endpoint: "{{ .Values.observe.collectionEndpoint | toString | trimSuffix "/" }}/v1/prometheus"
+    endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/prometheus"
     headers:
-        authorization: "Bearer {{ .Values.observe.token }}"
+        authorization: "${env:OBSERVE_TOKEN}"
     resource_to_telemetry_conversion:
         enabled: true # Convert resource attributes to metric labels
 {{- end -}}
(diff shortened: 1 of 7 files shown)


charts/agent/templates/_daemonset-logs-metrics-config.tpl
```diff
@@ -168,11 +168,11 @@ service:
         receivers: [filelog]
         processors: [memory_limiter, batch, resourcedetection/cloud, k8sattributes, attributes/observe_common, attributes/debug_source_pod_logs]
         exporters: [otlphttp/observe/base, debug/override]
-      hostmetrics:
+      metrics/hostmetrics:
```
xueweiz, line 171: ah sorry, just saw this...
Sorry for the bug!
I uploaded a PR before seeing your PR...https://github.com/observeinc/helm-charts/pull/192

charts/agent/templates/_config-exporters.tpl
```diff
@@ -2,21 +2,21 @@
 otlphttp/observe/base:
     endpoint: "{{ .Values.observe.collectionEndpoint | toString | trimSuffix "/" }}/v2/otel"
     headers:
-        authorization: "Bearer {{ .Values.observe.token }}"
```
xueweiz, line 5: Question: it seems that historically our agents does expects the "Bearer" string to be outside of the token [1]. But it seems you are expect that string in the token here? Is that intentional?

(And I feel like the env var name probably should be `TOKEN`, instead of `OBSERVE_TOKEN`, given how you set it up. But I might be wrong. If your local testing works, please ignore me :P

[1]
https://github.com/observeinc/helm-charts/blob/c62a44b5bdf29f2b9da1c63db8e9b31506ead36f/charts/traces/values.yaml#L106
https://github.com/observeinc/manifests/blob/a4261459b61fc4fe2e17684f3d5bfcbb8cd1ea88/bases/logs/m/fluent-bit.conf#L95

charts/agent/templates/_daemonset-logs-metrics-config.tpl
arthur-observe: ok - reverted commit

charts/agent/templates/_config-exporters.tpl
arthur-observe: After talking to @obs-gh-alexlew figured out that agent automatically does substitution  if TOKEN env var exists

arthur-observe: Also agent does Bearer portion as well

.github/workflows/pre-commit.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     uses: observeinc/.github/.github/workflows/shared_commit-validation.yaml@main
     secrets: inherit
     with:
-      skip: '{}'
+      skip: '{"jobs": ["single-commit"]}'
```
obs-gh-abhishekrao, line 13: PLEASE remove this skip. CI workflows depend on a single commit being present. Otherwise we CANNOT determine the nature of commit to version the chart!

obs-gh-abhishekrao (changes requested):

charts/agent/templates/secret.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if .Values.observe.token.create -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: agent-credentials
+type: Opaque
+data:
+  OBSERVE_TOKEN: {{ .Values.observe.token.value | b64enc | quote }}
+  ENTITY_TOKEN: {{ .Values.observe.entityToken.value | b64enc | quote }}
```
obs-gh-alexlew, line 9: what is an entity token? and they can provide it via this secret or via the values.yaml?

xueweiz: That is the token used to send data to the direct write entity table :)
(Eventually, we should have 3 tokens: one for turnkey logs, one for turnkey metrics, one for direct-write entity.

obs-gh-alexlew: why do we need a separate token? can't we just create the base datastream to be entity table enabled?

xueweiz: for posterity: we want it because: https://observeinc.slack.com/archives/C06RY0RQ7TP/p1724788582005679?thread_ts=1724784451.989259&cid=C06RY0RQ7TP

PR #193, opened 2024-08-27
feat: add updatecli configs for agent stack

obs-gh-alexlew:


obs-gh-alexlew:
seems like test-charts is just broken? specifically the `helm/chart-testing-action`

https://github.com/helm/chart-testing-action/issues/156 might fix?

obs-gh-abhishekrao:
Do we also need `updatecli/external` rules for when the agent has a new image or something similar? Currently it seems we refer to the opentelemetry-collector charts so that's another angle to look at as well.

obs-gh-alexlew:
to me it seems like bumping otel-collector-chart/agent versions should be a manual action and not something automated by updatecli. We don't always want to be on latest. It is useful to open a PR for us with updates but until the integration testing is ready, I think it's better to leave bumps + testing as a manual action.

obs-gh-abhishekrao (approved):

PR #194, opened 2024-08-27
feat: Add Workload entities with straightforward facets

obs-gh-enricogiorio:
Add a variety of facets for the following new entities:

- Job
- CronJob
- StatefulSet
- DaemonSet

### DIFF
--- a/charts/agent/templates/_deployment-cluster-events-config.tpl
+++ b/charts/agent/templates/_deployment-cluster-events-config.tpl
@@ -123,7 +123,7 @@ processors:
           - set(attributes["observe_transform"]["facets"]["deploymentName"], body["metadata"]["ownerReferences"][0]["name"]) where body["metadata"]["ownerReferences"][0]["kind"] == "ReplicaSet"
           - replace_pattern(attributes["observe_transform"]["facets"]["deploymentName"], "^(.*)-[0-9a-f]+$$", "$$1")
           - set(attributes["observe_transform"]["facets"]["deploymentName"], body["metadata"]["ownerReferences"][0]["name"]) where body["metadata"]["ownerReferences"][0]["kind"] == "Deployment"
-      # For Pod
+      # Pod
       - context: log
         conditions:
           - body["kind"] == "Pod"
@@ -135,13 +135,13 @@ processors:
           - set(attributes["observe_transform"]["facets"]["startTime"], body["status"]["startTime"])
           - set(attributes["observe_transform"]["facets"]["readinessGates"], body["object"]["spec"]["readinessGates"])
           - set(attributes["observe_transform"]["facets"]["nodeName"], body["spec"]["nodeName"])
-      # For Namespace
+      # Namespace
       - context: log
         conditions:
           - body["kind"] == "Namespace"
         statements:
           - set(attributes["observe_transform"]["facets"]["status"], body["status"]["phase"])
-      # For Node
+      # Node
       - context: log
         conditions:
           - body["kind"] == "Node"
@@ -151,7 +151,7 @@ processors:
           - set(attributes["observe_transform"]["facets"]["kubeletVersion"], body["status"]["nodeInfo"]["kubeletVersion"])
           - set(attributes["observe_transform"]["facets"]["osImage"], body["status"]["nodeInfo"]["osImage"])
           - set(attributes["observe_transform"]["facets"]["taints"], body["spec"]["taints"])
-      # For Deployment
+      # Deployment
       - context: log
         conditions:
           - body["kind"] == "Deployment"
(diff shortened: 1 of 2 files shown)


charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -186,6 +186,67 @@ processors:
...
+          - set(attributes["observe_transform"]["facets"]["updatedNumberScheduled"], body["status"]["updatedNumberScheduled"])
+          - set(attributes["observe_transform"]["facets"]["numberAvailable"], body["status"]["numberAvailable"])
+          - set(attributes["observe_transform"]["facets"]["numberUnavailable"], body["status"]["numberUnavailable"])
+          - set(attributes["observe_transform"]["facets"]["numberMisscheduled"], body["status"]["numberMisscheduled"])
+          # spec
+          - set(attributes["observe_transform"]["facets"]["updateStrategy"], body["spec"]["updateStrategy"]["type"])
+          - set(attributes["observe_transform"]["facets"]["maxUnavailable"], body["spec"]["updateStrategy"]["rollingUpdate"]["maxUnavailable"])
+          - set(attributes["observe_transform"]["facets"]["maxUnavailable"], 0) where attributes["observe_transform"]["facets"]["maxUnavailable"] == nil
+          - set(attributes["observe_transform"]["facets"]["maxSurge"], body["spec"]["updateStrategy"]["rollingUpdate"]["maxSurge"])
+          - set(attributes["observe_transform"]["facets"]["maxSurge"], 0) where attributes["observe_transform"]["facets"]["maxSurge"] == nil
```
xueweiz, line 234: curious, why do you set it to 0 when the field doesn't exist?
(i.e. is 0 better than nil? if so, under what situation, and why?

charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -186,6 +186,67 @@ processors:
...
+        conditions:
+          - body["kind"] == "StatefulSet"
+        statements:
+          # status
+          - set(attributes["observe_transform"]["facets"]["currentReplicas"], body["status"]["currentReplicas"])
+          - set(attributes["observe_transform"]["facets"]["readyReplicas"], body["status"]["readyReplicas"])
+          # spec
+          - set(attributes["observe_transform"]["facets"]["service"], body["spec"]["serviceName"])
+          - set(attributes["observe_transform"]["facets"]["podManagementPolicy"], body["spec"]["podManagementPolicy"])
+          - set(attributes["observe_transform"]["facets"]["replicas"], body["spec"]["replicas"])
```
xueweiz, line 246: desiredReplicas?
(that's what we call it for Deployment

charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -186,6 +186,67 @@ processors:
...
+          - set(attributes["observe_transform"]["facets"]["updateStrategy"], body["spec"]["updateStrategy"]["type"])
+          - set(attributes["observe_transform"]["facets"]["partition"], body["spec"]["updateStrategy"]["rollingUpdate"]["partition"])
+          - set(attributes["observe_transform"]["facets"]["partition"], 0) where attributes["observe_transform"]["facets"]["partition"] == nil
```
xueweiz, line 249: same as above, is 0 better than nil? For what use case?

charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -186,6 +186,67 @@ processors:
...
+          - set(attributes["observe_transform"]["facets"]["activePods"], body["status"]["active"])
+          - set(attributes["observe_transform"]["facets"]["failedPods"], body["status"]["falied"])
+          - set(attributes["observe_transform"]["facets"]["succeededPods"], body["status"]["succeeded"])
+          - set(attributes["observe_transform"]["facets"]["readyPods"], body["status"]["ready"])
+          # spec
+          - set(attributes["observe_transform"]["facets"]["completions"], body["spec"]["completions"])
+          - set(attributes["observe_transform"]["facets"]["parallelism"], body["spec"]["parallelism"])
+          - set(attributes["observe_transform"]["facets"]["activeDeadlineSeconds"], body["spec"]["activeDeadlineSeconds"])
+          - set(attributes["observe_transform"]["facets"]["backoffLimit"], body["spec"]["backoffLimit"])
+      # Cron Job
```
xueweiz, line 205: tiny nit: maybe just CronJob, DaemonSet

charts/agent/templates/_deployment-cluster-events-config.tpl
obs-gh-enricogiorio: I had it nil, but then looking at our code we are already doing this thing for readyReplicas (Deployment and ReplicaSet). I think `nil` is better tho. I will revert this and remove the set to 0.

xueweiz (approved):

PR #195, opened 2024-08-27
feat: OB-35984 add conditional enabled values for each deployment

arthur-observe:
- add conditions to chart.yaml and values.yaml to provide override for each deployment

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -1,8 +1,5 @@
-
-namespaceOverride: &namespaceOverride "observe"
-
 observe:
-  #If create = false it is assumed that a secret named agent-credentials with key: OBSERVE_TOKEN already exists
+  # If create = false it is assumed that a secret named agent-credentials with key: OBSERVE_TOKEN already exists
   token:
     create: true
     value: ""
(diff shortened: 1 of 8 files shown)


xueweiz (approved):

charts/agent/values.yaml
```diff
@@ -36,6 +36,7 @@ config:
 
 ################################################
 deployment-cluster-events:
+  enabled: true
```
obs-gh-alexlew, line 39: i think these should go into the top level config vs. in here. also, im curious about the `config` top level field, isn't this all "config"? 

I think this field specifically should be `cluster.entities.enabled` or `cluster.resources.enabled`

charts/agent/values.yaml
```diff
@@ -150,6 +151,7 @@ deployment-cluster-events:
 
 ################################################
 deployment-cluster-metrics:
+  enabled: true
```
obs-gh-alexlew, line 154: this field should be `cluster.metrics.enabled`

charts/agent/values.yaml
```diff
@@ -262,6 +264,7 @@ deployment-cluster-metrics:
 
 ################################################
 daemonset-logs-metrics:
+  enabled: true
```
obs-gh-alexlew, line 267: really this should be split into 2 configs, `containers.logs.enabled` and `containers.metrics.enabled`

charts/agent/values.yaml
```diff
@@ -406,6 +409,7 @@ daemonset-logs-metrics:
 
 ################################################
 deployment-agent-monitor:
+  enabled: true
```
obs-gh-alexlew, line 412: maybe this is just a top level config `agent_self_monitoring.enabled`

PR #198, opened 2024-08-29
feat: OB-35984 add optional env vars for tokens

arthur-observe:
- Made TOKEN and ENTITY_TOKEN environment variables optional and adjusted template files to accomodate

### DIFF
--- a/charts/agent/templates/_deployment-cluster-events-config.tpl
+++ b/charts/agent/templates/_deployment-cluster-events-config.tpl
@@ -5,7 +5,7 @@ extensions:
 
 exporters:
 {{- include "config.exporters.debug" . | nindent 2 }}
-{{ if .Values.observe.entityToken }}
+{{ if .Values.observe.entityToken.create }}
 {{- include "config.exporters.otlphttp.observe.entity" . | nindent 2 }}
 {{ end }}
 {{- include "config.exporters.otlphttp.observe.base" . | nindent 2 }}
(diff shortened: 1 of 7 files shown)


charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -5,7 +5,7 @@ extensions:
 
 exporters:
 {{- include "config.exporters.debug" . | nindent 2 }}
-{{ if .Values.observe.entityToken }}
+{{ if .Values.observe.entityToken.create }}
 {{- include "config.exporters.otlphttp.observe.entity" . | nindent 2 }}
 {{ end }}
 {{- include "config.exporters.otlphttp.observe.base" . | nindent 2 }}
```
xueweiz, line 11: Should we wrap this in a `{{ if .Values.observe.token.create }}`?
I don't think we absolutely need this, so if this doesn't work, I don't mind if we just leave it as is.
But sounds like the conditional block is the right thing to do?

(Same can be said for the other deployment/daemonset's exporters. If they depends on `observe.token`, maybe only enable them when the token is there.

charts/agent/templates/secret.yaml
```diff
@@ -1,9 +1,14 @@
-{{- if .Values.observe.token.create -}}
+{{- if or .Values.observe.token.create .Values.observe.entityToken.create -}}
 apiVersion: v1
 kind: Secret
 metadata:
   name: agent-credentials
 type: Opaque
 data:
+  {{- if or .Values.observe.token.create }}
```
obs-gh-alexlew, line 8: whats the or value here?

obs-gh-alexlew (approved):

PR #202, opened 2024-08-30
feat: Add facets for several new resources

obs-gh-enricogiorio:
This facet was computed using OTTL's Len() on a slice of pointers, which is not a supported ValueType for Len().

Therefore, this facet will have to be computed with a custom processor.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.8.1
+version: 0.9.0
 appVersion: "1.0.0"
 dependencies:
   - name: opentelemetry-collector
--- a/charts/agent/templates/_deployment-cluster-events-config.tpl
+++ b/charts/agent/templates/_deployment-cluster-events-config.tpl
@@ -129,10 +129,13 @@ processors:
           - body["kind"] == "Pod"
         statements:
           - set(attributes["observe_transform"]["identifiers"]["podName"], body["metadata"]["name"])
+          - set(attributes["observe_transform"]["identifiers"]["podUid"], body["metadata"]["uid"])
+          # status
           - set(attributes["observe_transform"]["facets"]["phase"], body["status"]["phase"])
           - set(attributes["observe_transform"]["facets"]["podIP"], body["status"]["podIP"])
           - set(attributes["observe_transform"]["facets"]["qosClass"], body["status"]["qosClass"])
           - set(attributes["observe_transform"]["facets"]["startTime"], body["status"]["startTime"])
+          # spec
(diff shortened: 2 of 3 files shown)


charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -235,11 +254,13 @@ processors:
...
           - body["kind"] == "StatefulSet"
         statements:
+          - set(attributes["observe_transform"]["identifiers"]["statefulSetName"], body["metadata"]["name"])
+          - set(attributes["observe_transform"]["identifiers"]["statefulSetUid"], body["metadata"]["uid"])
           # status
           - set(attributes["observe_transform"]["facets"]["currentReplicas"], body["status"]["currentReplicas"])
           - set(attributes["observe_transform"]["facets"]["readyReplicas"], body["status"]["readyReplicas"])
           # spec
-          - set(attributes["observe_transform"]["facets"]["service"], body["spec"]["serviceName"])
+          - set(attributes["observe_transform"]["facets"]["serviceName"], body["spec"]["serviceName"])
```
xueweiz, line 263: Good catch!

charts/agent/Chart.yaml
```diff
@@ -2,30 +2,30 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.8.0
+version: 0.8.1
 appVersion: "1.0.0"
 dependencies:
-  - name: opentelemetry-collector
```
xueweiz, line 8: the space edits seems weird, maybe your linter is different?

obs-gh-enricogiorio: Yeah I found a weird setting in my formatter that makes lists not indented. Weird.

arthur-observe (changes requested):
There should be a Chart.lock update

if you run make lint from root then pre-commit run -a (twice) should get everything right

charts/agent/templates/_deployment-cluster-events-config.tpl
```diff
@@ -226,24 +251,127 @@ processors:
...
+          - set(attributes["observe_transform"]["facets"]["type"], body["type"])
+      - context: log
+        conditions:
+          - body["kind"] == "PersistentVolume"
+        statements:
+          - set(attributes["observe_transform"]["identifiers"]["persistentVolumeName"], body["metadata"]["name"])
+          - set(attributes["observe_transform"]["identifiers"]["persistentVolumeUid"], body["metadata"]["uid"])
+          # spec
+          - set(attributes["observe_transform"]["facets"]["capacity"], body["spec"]["capacity"]["storage"])
+          - set(attributes["observe_transform"]["facets"]["persistentVolumeReclaimPolicy"], body["spec"]["persistentVolumeReclaimPolicy"])
```
colecorrente, line 295: for these facets, why prefix with "persistentVolume" 

ie why not just "volumeReclaimPolicy" and "volumeClass"


same for PersistentVolumeClaim below

PR #207, opened 2024-09-05
chore: Remove helm-docs footer

obs-gh-enricogiorio:
On one hand, it is not possible to generate the helm-docs version footer when helm-docs is installed via go install, which is the most common and easiest way to install this tool.

On the other hand, configuring helm-docs (when installed via brew or other package managers) to skip printing its version footer is very easy (passing --skip-version-footer).

Make skipping the version footer the default behavior to facilitate development when helm-docs is installed via go install.

More context - https://github.com/norwoodj/helm-docs/pull/220


obs-gh-alexlew (approved):

obs-gh-abhishekrao (approved):
LGTM. I wish update-cli rules were less aggressive so a PR like this wouldn't require chart version bumping, but that's for a future improvement.

PR #219, opened 2024-09-16
feat: split container logs and metrics configs

obs-gh-alexlew:


charts/agent/README.md
```diff
@@ -39,6 +39,7 @@ Chart to install K8s collection stack based on Observe Agent
 | cluster.name | string | `"observe-agent-monitored-cluster"` |  |
 | cluster.namespaceOverride.value | string | `"observe"` |  |
 | cluster.uidOverride.value | string | `""` |  |
+| containers.enabled | bool | `true` |  |
```
obs-gh-alexlew, line 42: had to add this new top level field since the `Chart.yaml dependencies.condition` doesn't allow you to do operators (like `if`) or to specify multiple fields so we need one field to represent whether we should create the daemonset or not.

PR #221, opened 2024-09-17
ci: add addnl integration tests

obs-gh-nikhildua:
Addresses https://observe.atlassian.net/browse/OB-36071

- Add advanced integration tests
- Allows framework for matrixing of namespace & values files for non-default helm configuration testing


integration/modules/deploy_helm/values/default.yaml
```diff
@@ -1,8 +1,18 @@
...
+#Common for namepsace/url/token
 observe:
-    collectionEndpoint:
-      value: ${observe_url}
-    token:
-      value: ${observe_token}
+  collectionEndpoint:
+    value: ${observe_url}
+  token:
+    value: ${observe_token}
```
obs-gh-alexlew, line 15: This schema is changing pretty rapidly but there's a new top level object `node` as well now

obs-gh-alexlew (approved):

integration/modules/deploy_helm/values/default.yaml
obs-gh-nikhildua: @obs-gh-alexlew are you referring to this? 
https://github.com/observeinc/helm-charts/blob/main/charts/agent/values.yaml#L32 

If so, this "default" file is just meant to overwrite anything related to namespace / token / endpoints

PR #224, opened 2024-09-19
feat: rename all services to be simpler

obs-gh-alexlew:


obs-gh-nikhildua:
Integration testing seems fine with the name change.

xueweiz (approved):

PR #226, opened 2024-09-20
feat: simplify Event's identifiers field

xueweiz:
Nested objects are harder to handle, let's avoid that, and focus on the most useful information here.

### DIFF
--- a/charts/agent/templates/_cluster-events-config.tpl
+++ b/charts/agent/templates/_cluster-events-config.tpl
@@ -195,7 +195,9 @@ processors:
         conditions:
           - body["kind"] == "Event"
         statements:
-          - set(attributes["observe_transform"]["identifiers"]["involvedObject"], body["involvedObject"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectKind"], body["involvedObject"]["kind"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectName"], body["involvedObject"]["name"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectUid"], body["involvedObject"]["uid"])
           - set(attributes["observe_transform"]["facets"]["firstTimestamp"], body["firstTimestamp"])
           - set(attributes["observe_transform"]["facets"]["lastTimestamp"], body["lastTimestamp"])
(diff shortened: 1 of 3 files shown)


charts/agent/templates/_cluster-events-config.tpl
```diff
@@ -195,7 +195,9 @@ processors:
         conditions:
           - body["kind"] == "Event"
         statements:
-          - set(attributes["observe_transform"]["identifiers"]["involvedObject"], body["involvedObject"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectKind"], body["involvedObject"]["kind"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectName"], body["involvedObject"]["name"])
+          - set(attributes["observe_transform"]["identifiers"]["involvedObjectUid"], body["involvedObject"]["uid"])
```
obs-gh-jabernet, line 200: Shouldn't the UID be enough for the hashing? Or are the other fields there for other purposes?

xueweiz: We include these fields not to build the hashing (the identifiers hash is an optimization, it's not a goal / use case).
We include these fields so FE can easily find related events for an object.

For example, in a pod instance page, we'd need to find the list of events associated with that pod, which the query would minimally look like:
```
filter kind = "Event"
filter identifiers.involvedObjectUid = "1234-3546"
statsby facets: last(facets), group_by(identifiers)
```

Technically, yes, for a given object, on the instance page since we have its Uid, we can build this query with purely a Uid filter.
However, I generally try to preserve some object kind and name information as well, in case we have use cases where we only have the `clusterName, namespaceName, objectName` as the object identifier [1]. For those, we will construct the filter using those.

These data should be small enough :) I don't think we are that tight on the a few bytes here :D 


[1] In general, we can either use `clusterName, namespaceName, objectName` as ID, or `clusterName, objectUid` as ID. The latter is stronger. But a lot of times the former is more convenience, more available, and more readable. So, I think it's better to offer both approach here.

obs-gh-jabernet: Make sense, thanks for the detailed explanation, that's what I wanted to understand.

xueweiz:
Fixed the readme in the latest push

PR #232, opened 2024-09-30
feat: OB-36454 add pod metrics config

arthur-observe:
- Add configuration for pod metrics to: 
  - values.yaml 
  - _cluster-metrics-config.tpl
- Add example for pod metrics to helm-charts/examples/agent/pod_metrics

### DIFF
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.17.0](https://img.shields.io/badge/Version-0.17.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
+![Version: 0.18.0](https://img.shields.io/badge/Version-0.18.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
 
 > [!CAUTION]
 > This chart is under active development and is not meant to be installed yet.
@@ -166,6 +166,14 @@ Chart to install K8s collection stack based on Observe Agent
 | cluster.events.enabled | bool | `true` |  |
 | cluster.events.pullInterval | string | `"20m"` |  |
 | cluster.metrics.enabled | bool | `true` |  |
+| cluster.metrics.pod.action | string | `"keep"` |  |
+| cluster.metrics.pod.enabled | bool | `false` |  |
+| cluster.metrics.pod.interval | string | `"10s"` |  |
+| cluster.metrics.pod.metric_drop_regex | string | `".*bucket"` |  |
+| cluster.metrics.pod.metric_keep_regex | string | `"(.*)"` |  |
+| cluster.metrics.pod.namespace_drop_regex | string | `"(.*istio.*|.*ingress.*|kube-system|observe)"` |  |
+| cluster.metrics.pod.namespace_keep_regex | string | `"(.*)"` |  |
+| cluster.metrics.pod.port_keep_regex | string | `".*metrics"` |  |
 | cluster.name | string | `"observe-agent-monitored-cluster"` |  |
 | cluster.namespaceOverride.value | string | `"observe"` |  |
 | cluster.uidOverride.value | string | `""` |  |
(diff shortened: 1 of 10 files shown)


charts/agent/templates/_cluster-metrics-config.tpl
```diff
@@ -23,6 +23,109 @@ receivers:
...
+          - action: 'keep'
+            source_labels: ['__meta_kubernetes_namespace']
+            regex: {{.Values.cluster.metrics.pod.namespace_keep_regex}}
+
+          # Drop endpoints without one of: a port name suffixed with the configured regex, or an explicit prometheus port annotation.
+          - action: 'keep'
+            source_labels: ['__meta_kubernetes_pod_container_port_name', '__meta_kubernetes_pod_annotation_prometheus_io_port']
+            regex: '({{.Values.cluster.metrics.pod.port_keep_regex}};|.*;\d+)'
+
+          # Drop pods with phase Succeeded or Failed.
```
obs-gh-alexlew, line 54: why do these get dropped? no scraping metrics from job pods basically?

charts/agent/values.yaml
```diff
@@ -21,6 +21,16 @@ cluster:
     enabled: true
   metrics:
     enabled: true
+    pod:
+      enabled: false
+      action: keep
```
obs-gh-alexlew, line 26: what are the valid values here? and what does this do?

charts/agent/values.yaml
```diff
@@ -21,6 +21,16 @@ cluster:
...
+      interval: 10s
+      # add whatever namespace the collection is deployed to
+      namespace_drop_regex: (.*istio.*|.*ingress.*|kube-system|observe)
+      namespace_keep_regex: (.*)
```
obs-gh-alexlew, line 30: how does this interact with namespace_drop_regex?

charts/agent/templates/_cluster-metrics-config.tpl
arthur-observe: This is same as existing - I assume it will just keep scraping same metrics?  I don't actually know.

charts/agent/values.yaml
arthur-observe: enabled - true/false
action - keep / drop

charts/agent/README.md
```diff
@@ -166,6 +166,14 @@ Chart to install K8s collection stack based on Observe Agent
 | cluster.events.enabled | bool | `true` |  |
 | cluster.events.pullInterval | string | `"20m"` |  |
 | cluster.metrics.enabled | bool | `true` |  |
+| cluster.metrics.pod.action | string | `"keep"` |  |
```
obs-gh-alexlew, line 169: I'm actually not sure this even belongs nested under cluster at all. might make sense to break this out to a new top level section called `application`

charts/agent/values.yaml
arthur-observe: they act independently

examples/agent/affinity/affinity-values.yaml
```diff
@@ -47,15 +47,14 @@ deployment-cluster-events:
                 operator: NotIn
                 values: [windows]
 
-deployment-cluster-metrics:
```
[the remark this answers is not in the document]
arthur-observe: that's examples directory - guessing none cleaned that up

charts/agent/values.yaml
```diff
@@ -21,6 +21,16 @@ cluster:
     enabled: true
   metrics:
     enabled: true
+    pod:
+      enabled: false
+      action: keep
+      interval: 10s
+      # add whatever namespace the collection is deployed to
+      namespace_drop_regex: (.*istio.*|.*ingress.*|kube-system|observe)
```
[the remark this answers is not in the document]
arthur-observe: that follows our existing helm chart approach - happy to rename them

charts/agent/README.md
arthur-observe: put it there because thats where the other switches for that container are - no quarrels with moving it out - we'll have to make sure we deal with enabled switch in chart but that should be np

charts/agent/values.yaml
arthur-observe: these flags end up having same effect - action is the way it was gated in grafana agent - I left it in but we should probably just delete it

xueweiz (approved):

charts/agent/values.yaml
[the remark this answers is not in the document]
obs-gh-alexlew: hopefully people won't really need to experience more than one of our helm charts in the future, i think its okay for us to be inconsistent here and stick to published helm best practices

PR #233, opened 2024-10-01
feat: OB-36454 update values to new conventions and add and update tests

arthur-observe:
- Update values file to give more descriptive names and add comments for clarity
- Add new examples for prometheus labels

### DIFF
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.18.0](https://img.shields.io/badge/Version-0.18.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
+![Version: 0.19.0](https://img.shields.io/badge/Version-0.19.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
 
 > [!CAUTION]
 > This chart is under active development and is not meant to be installed yet.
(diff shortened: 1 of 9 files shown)


charts/agent/README.md
```diff
@@ -169,11 +169,11 @@ Chart to install K8s collection stack based on Observe Agent
 | cluster.metrics.pod.action | string | `"keep"` |  |
```
obs-gh-alexlew, line 169: is this used anywhere then if we're defaulting to keep?

NathanNam (approved):
LGTM

charts/agent/README.md
```diff
@@ -169,11 +169,11 @@ Chart to install K8s collection stack based on Observe Agent
...
 | cluster.metrics.pod.enabled | bool | `false` |  |
 | cluster.metrics.pod.interval | string | `"10s"` |  |
-| cluster.metrics.pod.metric_drop_regex | string | `".*bucket"` |  |
-| cluster.metrics.pod.metric_keep_regex | string | `"(.*)"` |  |
-| cluster.metrics.pod.namespace_drop_regex | string | `"(.*istio.*|.*ingress.*|kube-system|observe)"` |  |
-| cluster.metrics.pod.namespace_keep_regex | string | `"(.*)"` |  |
-| cluster.metrics.pod.port_keep_regex | string | `".*metrics"` |  |
+| cluster.metrics.pod.metricDropRegex | string | `".*bucket"` |  |
```
obs-gh-alexlew, line 172: re: these being nested under cluster, we don't need the top level groups to match exactly with the services we're actually deploying since these values just get used downstream in the actual configs as flags. I think its better to organize top level by use case/category of data in which case `application` makes more sense as a standalone. In the future if we want to have some configurations around traces at this level too it would make sense under `application` moreso than cluster

obs-gh-alexlew: names look good to me though, i do kind of want some sort of comment explaining how the drop and keep interact. what happens if i define rules that collide? like i want to keep everything but then drop some things? drop takes precedent I assume? would love to have that written explicitly

charts/agent/README.md
arthur-observe: that option is gone now - seeing as we have enable as a value it is set to keep all the time in config and it just won't be added if we don't set enabled = true

charts/agent/README.md
arthur-observe: done

obs-gh-alexlew (approved):

PR #234, opened 2024-10-01
feat: Add overrides for excludeTargets and includeTargets

obs-gh-justindaines:
Adding option to override [kube-state-events](https://github.com/observeinc/kube-state-events/tree/main) container with [include-target](https://github.com/observeinc/kube-state-events/blob/6d6b1d1e1fa39f843db75e387b1d7126a68fb132/cmd/kube-state-events/main.go#L111) and [exclude-target](https://github.com/observeinc/kube-state-events/blob/6d6b1d1e1fa39f843db75e387b1d7126a68fb132/cmd/kube-state-events/main.go#L116).

Validated with following test values:
```
containerOverrides:
  excludeTargets:
    - events.k8s.io/v1/events
    - apps/v1/replicasets
  includeTargets:
    - apps/v1/deployments
```

### DIFF
--- a/charts/events/templates/deployment.yaml
+++ b/charts/events/templates/deployment.yaml
@@ -47,6 +47,18 @@ spec:
           image: {{ .Values.image.kube_state_events.repository }}:{{ default .Chart.AppVersion .Values.image.kube_state_events.tag }}
           imagePullPolicy: {{ .Values.image.kube_state_events.pullPolicy }}
           args:
+            # Include excludeTags if defined
+            {{- if .Values.containerOverrides.excludeTargets }}
+            {{- range .Values.containerOverrides.excludeTargets }}
+            - -exclude-target={{ . }}
+            {{- end }}
+            {{- end }}
+            # Include includeTags if defined
+            {{- if .Values.containerOverrides.includeTargets }}
+            {{- range .Values.containerOverrides.includeTargets }}
+            - -include-target={{ . }}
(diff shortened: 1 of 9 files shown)


charts/events/templates/deployment.yaml
```diff
@@ -55,6 +67,7 @@ spec:
             - autoscaling/v1
```
obs-gh-abhishekrao, line 67: What happens when either of these three are specified in `-exclude-target` or `-include-target`?

obs-gh-justindaines: Adding `-include-target` doesn't add any value. I included it for possible future use. Do you think we should drop it?

I did catch an error, should be [] and not {}. I'll update accordingly, but would appreciated thoughts on omitting `includeTargets`.

obs-gh-abhishekrao: I think it's good to have that, but my question was different. It looks like additional APIs can be included by simply specifying `-<API>`, which you can see specified as `autoscaling/v1`, `batch/v1`, `networking.k8s.io/v1`. I was curious as to how the new flags behave with these in tandem.

obs-gh-abhishekrao:
Also, please see if you can add these flags to the default CI tests in `charts/stack/ci/test-values.yaml`. That'll help capture if this change introduces any regressions or works as expected.

obs-gh-abhishekrao (approved):

PR #237, opened 2024-10-03
feat: OB-37318 Add container ID attribute

obs-gh-enricogiorio:
Collect container id for both metrics and logs.

 - metrics: use kubeletstats receiver's "extra_metadata_labels"
 - logs: use resource processor to take the container ID from the log.file.path (attribute added by the filelog receiver)

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -29,6 +29,7 @@ k8sattributes:
     - k8s.node.name
     - k8s.node.uid
     - k8s.container.name
+    - container.id
   passthrough: false
   pod_association:
   - sources:
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.26.0
-appVersion: "1.3.0"
+version: 0.27.0
+appVersion: "1.1.0"
 dependencies:
   - name: opentelemetry-collector
     version: 0.101.1
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.26.0](https://img.shields.io/badge/Version-0.26.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.3.0](https://img.shields.io/badge/AppVersion-1.3.0-informational?style=flat-square)
+![Version: 0.27.0](https://img.shields.io/badge/Version-0.27.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
 
 > [!CAUTION]
 > This chart is under active development and is not meant to be installed yet.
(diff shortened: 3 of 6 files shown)


charts/agent/templates/_config-processors.tpl
```diff
@@ -81,3 +81,24 @@ memory_limiter:
...
 observek8sattributes:
 {{- end -}}
+
+{{- define "config.processors.transform.observe_common"}}
+transform/observe_common:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+        # This is tricky, but essentially context:resource ensures that the
```
xueweiz, line 91: I wonder if this might be an overkill.
I just checked `k8sattributes` processor, seems it can report `container.id`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/documentation.md

It's just not enabled by default. Can we try enabling that?
(And, if it works, we might as well use that for both logs and metrics.

obs-gh-enricogiorio: I saw that, but it says that it requires `k8s.container.restart_count` to guarantee the ID is correct, which is not available and it's not easy to get. One way is to expose it directly from the configs, but we obviously cannot rely on that.
[Others have the same problem](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/16431), and eventually you need to do the same "filename hack" to get the restart count so that k8sattributes can get the container id 😂 .
So we can just do the hack to get the ID directly I think

xueweiz: > which is not available

Are you sure `k8s.container.restart_count` is not available? :)
checkout this worksheet's `resource_attributes` column :D 

https://102.observe-o2.com/workspace/41000009/log-explorer?datasetId=41051312&requiredCorrelationTags=k8s.pod.name&requiredCorrelationTags=k8s.cluster.name&requiredCorrelationTags=k8s.namespace.name&requiredCorrelationTags=k8s.pod.uid&s=1083-n5n1i77u

I guess the information is provided by OTel collector's built in log file processor, try follow this: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/beb7e34fd7c7ac08f5a9f2dae1f8fd618b50c69a/pkg/stanza/operator/parser/container/parser.go#L31

I checked, I couldn't find a single log line where we have a `k8s.container.name` without `k8s.container.restart_count`: https://102.observe-o2.com/workspace/41000009/log-explorer?datasetId=41051312&requiredCorrelationTags=k8s.pod.name&requiredCorrelationTags=k8s.cluster.name&requiredCorrelationTags=k8s.namespace.name&requiredCorrelationTags=k8s.pod.uid&s=1083-c4o13kaa

So, I think we can rely on it being there :)

obs-gh-enricogiorio: I see. I tried looking and I have no idea where this is set or which receiver/processor populates it. But indeed it's there in O2 in all logs ahah.
I will add it then.

obs-gh-enricogiorio: This is the row above. No `container.id` extracted despite matching with `restart_count` is possible.
image

obs-gh-enricogiorio:
image
As you can see, the only logs that don't have container.id are from namespace `kube-system`.

I will try to do [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md#cluster-scoped-rbac) to see if we can add RBAC permissions to the processor so we can enrich those as well!

xueweiz (approved):

PR #238, opened 2024-10-05
feat: OB-37317 add config options to cluster metrics

arthur-observe:
- Default metric collection interval to 60s for all metrics collection with interval option in values file and related configs
- Added rootPath and excludeMountPoints config for host metrics
- Add comments for clarity to various section of values and template files
- Add sending metadata and write ahead log (wal) config
- Add default empty tolerations option for pods

### DIFF
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.20.3](https://img.shields.io/badge/Version-0.20.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
+![Version: 0.21.0](https://img.shields.io/badge/Version-0.21.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.1.0](https://img.shields.io/badge/AppVersion-1.1.0-informational?style=flat-square)
 
 > [!CAUTION]
 > This chart is under active development and is not meant to be installed yet.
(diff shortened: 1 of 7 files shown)


charts/agent/README.md
```diff
@@ -342,8 +347,12 @@ Chart to install K8s collection stack based on Observe Agent
 | node.containers.logs.retryOnFailure.maxInterval | string | `"30s"` |  |
 | node.containers.logs.startAt | string | `"end"` |  |
 | node.containers.metrics.enabled | bool | `true` |  |
+| node.containers.metrics.interval | string | `"60s"` |  |
 | node.enabled | bool | `true` |  |
 | node.metrics.enabled | bool | `true` |  |
+| node.metrics.fileSystem.excludeMountPoints | string | `"[\"/dev/*\",\"/proc/*\",\"/sys/*\",\"/run/k3s/containerd/*\",\"/var/lib/docker/*\",\"/var/lib/kubelet/*\",\"/snap/*\"]"` |  |
```
obs-gh-alexlew, line 353: excludeMountPaths

charts/agent/README.md
```diff
@@ -342,8 +347,12 @@ Chart to install K8s collection stack based on Observe Agent
...
+| node.metrics.fileSystem.rootPath | string | `"/hostfs"` |  |
```
obs-gh-alexlew, line 354: what does this depend on?

charts/agent/values.yaml
```diff
@@ -73,9 +70,31 @@ node:
...
       startAt: end
-
+    # pulls node, pod, and container metrics from the API server on a kubelet and sends it down the metric pipeline for further processing.
     metrics:
       enabled: true
+      interval: 60s
+
+application:
+  # use this option to scrape prometheus metrics from pods
+  # !!! IMPORTANT !!! cluster.metrics.enabled must be set to true
```
obs-gh-alexlew, line 80: can we just do an enabled field here too and then make the if statement `cluster.metrics.enabled or application.enabled`?

charts/agent/README.md
arthur-observe: This was a list I picked up from the otel presets results.  It's been in the background the whole time - was bringing it forward as it seems like a control that will be good to make configurable.

charts/agent/values.yaml
arthur-observe: Helm chart doesn't sort this logic and the daemon set is conditional on cluster.metrics.enabled

I think for now we live with it until we find a better way

obs-gh-alexlew (approved):

PR #255, opened 2024-11-01
fix: Swap custom and transform processors in pipeline

obs-gh-enricogiorio:
The reason is that the custom processor might edit the log body in-place and the transform processor will copy parts of the body over to the facets. Since the transform processor is more flexible, we want to edit things first and eventually add them as facets once they are modified.

Example: secrets obfuscation. We want to redact the secrets' values that can show up in the annotations. The transform processor blindly copies the annotations to facets. So we don't want the secrets to show up in the facets un-redacted.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.30.3
+version: 0.30.4
 appVersion: "1.1.0"
 dependencies:
   - name: opentelemetry-collector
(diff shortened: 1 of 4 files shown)


xueweiz (approved):

obs-gh-enricogiorio:
This will only work once we release the agent with the "structured log body" patch.
The custom processor that we are moving before the `transform` processor is currently overriding the log body with a "flat" string, which precents the transform processor from indexing into it and do its thing.
I have a patch that makes the custom processor override the log body with a structured `map[string]any`, instead, which will allow the transform processor to do things on top of it.
Once that's merged, we will make a new public agent release, and then merge this fix here!

obs-gh-enricogiorio:
image
Local testing: my cluster sends Secret entities with obfuscated secrets' values in the facets after swapping processors.

PR #256, opened 2024-11-01
feat: Cleanup values.yaml with sensible defaults

obs-gh-enricogiorio:
Send everything to the /v1/kubernetes/v1/entity path

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -1,10 +1,8 @@
 # This section is for configuring agent to send data to Observe endpoints
-# we recommend using --set token.value=YOUR_TOKEN for values you want to keep secret
-# supplying a token creates a secret on your cluster
 observe:
   # If create = false it is assumed that a secret named agent-credentials with key: OBSERVE_TOKEN already exists
   token:
-    create: true
+    create: false
     value: ""
   collectionEndpoint:
     # ex -  https://12345678.collect.observeinc.com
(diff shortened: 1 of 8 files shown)


obs-gh-nikhildua:
https://github.com/observeinc/helm-charts/tree/main/integration/modules/deploy_helm/values
These are the values file that the integration tests refers to

Entity token is not specified for testing so it'll refer to whatever is in the default values file. Do we now need one?

obs-gh-enricogiorio:
I don't think we need one, I need to look into this very carefully. Will get back to you tomorrow

obs-gh-enricogiorio:
I see that the values are parametric themselves, i.e. they refer to stuff like {$observe_token}. Where are these variables defined?
It's definitely a test configuration issue since the patch changes the defaults and we might need to change the overrides as well.

charts/agent/values.yaml
```diff
@@ -152,9 +150,7 @@ cluster-events:
 
   image:
     repository: observeinc/observe-agent
-    # Overrides the image tag whose default is the chart appVersion.
-    # The service's name will be applied to the end of this value.
-    tag: "1.3.0"
+    tag: latest
```
xueweiz, line 153: Hum... Not pinning the image is likely too dangerous to be used as default. As user could see their agent "secretively" updating. Why do we need this as default?

xueweiz (approved):
LGTM! But you need to regenerate some files to pass the presubmit checks I think

xueweiz (approved):

PR #262, opened 2024-11-06
feat: Add key for fetching ingress class

obs-gh-enricogiorio:
Try fetching ingress class from the deprecated annotation "kubernetes.io/ingress.class", which might still contain the ingress class for outdated Ingress manifests.

### DIFF
--- a/charts/agent/templates/_cluster-events-config.tpl
+++ b/charts/agent/templates/_cluster-events-config.tpl
@@ -350,6 +350,7 @@ processors:
           - set(attributes["observe_transform"]["identifiers"]["ingressUid"], body["metadata"]["uid"])
           # spec
           - set(attributes["observe_transform"]["facets"]["class"], body["spec"]["ingressClassName"])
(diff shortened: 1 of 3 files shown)


charts/agent/templates/_cluster-events-config.tpl
```diff
@@ -353,6 +353,7 @@ processors:
           - set(attributes["observe_transform"]["identifiers"]["ingressUid"], body["metadata"]["uid"])
           # spec
           - set(attributes["observe_transform"]["facets"]["class"], body["spec"]["ingressClassName"])
+          - set(attributes["observe_transform"]["facets"]["class"], body["metadata"]["attributes"]["kubernetes.io/ingress.class"]) where attributes["observe_transform"]["facets"]["class"] == ""
```
xueweiz, line 356: I think you mean `"annotations"`?

obs-gh-enricogiorio: lol right

xueweiz (approved):

PR #266, opened 2024-11-07
fix: entity exporter should use logs_endpoint, instead of endpoint

xueweiz:
If you use endpoint, OTel agent will append it with "/logs" as the real URL. Not what we want.

### DIFF
--- a/charts/agent/templates/_config-exporters.tpl
+++ b/charts/agent/templates/_config-exporters.tpl
@@ -15,7 +15,7 @@ otlphttp/observe/base:
 
 {{- define "config.exporters.otlphttp.observe.entity" -}}
 otlphttp/observe/entity:
-    endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
+    logs_endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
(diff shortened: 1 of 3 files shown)


charts/agent/templates/_config-exporters.tpl
```diff
@@ -15,7 +15,7 @@ otlphttp/observe/base:
 
 {{- define "config.exporters.otlphttp.observe.entity" -}}
 otlphttp/observe/entity:
-    endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
+    logs_endpoint: "{{ .Values.observe.collectionEndpoint.value | toString | trimSuffix "/" }}/v1/kubernetes/v1/entity"
```
obs-gh-mattcotter, line 18: Do we also need `metrics_endpoint`/`traces_endpoint`?

xueweiz: (we don't support traces yet
for metrics, we use a dedicated prometheus endpoint + prometheus remotewrite exporter for it, and it has been running healthily for a while :) https://github.com/observeinc/helm-charts/blob/3c11bc01d8b181a5d86da943e5b3c104e6ea198b/charts/agent/templates/_config-exporters.tpl#L37

The prom remote write exporter does NOT do these "url suffixing" stuff :D

obs-gh-mattcotter (approved):

PR #274, opened 2024-11-27
feat: Populate metric unit of cpu metrics

obs-gh-enricogiorio:
Set the unit to "CPU" when the metric name contains *cpu* and the unit is not already set.

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -87,3 +87,13 @@ memory_limiter:
 {{- define "config.processors.attributes.observek8sattributes" -}}
 observek8sattributes:
 {{- end -}}
+
+{{- define "config.processors.transform.metrics" -}}
+transform/metrics:
+  metric_statements:
+    - context: metric
+      statements:
+        - set(unit, "test")
+        # - set(unit, "cores") where unit == ""
+        # - set(unit, "cores") where IsMatch(name, ".*cpu.*")
+{{- end -}}
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.30.6
+version: 0.31.0
 appVersion: "1.1.0"
 dependencies:
   - name: opentelemetry-collector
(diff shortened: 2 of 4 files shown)


charts/agent/templates/_config-processors.tpl
```diff
@@ -87,3 +87,13 @@ memory_limiter:
 {{- define "config.processors.attributes.observek8sattributes" -}}
 observek8sattributes:
 {{- end -}}
+
+{{- define "config.processors.transform.metrics" -}}
+transform/metrics:
+  metric_statements:
+    - context: metric
+      statements:
+        - set(unit, "test")
```
obs-gh-enricogiorio, line 96: These are my various attempts at making the change show up in o2. All of them have failed so far.
The [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/contexts/ottlmetric) say that I can access the "unit" field from the "metric" context, but apparently it's not the case, unless I am missing something.

obs-gh-alexlew: for some reason even the [metrics transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/metricstransformprocessor/README.md) doesn't mention being able to modify a metric's unit. i wonder if its prohibited for some reason

obs-gh-alexlew: found this example 
https://opentelemetry.io/blog/2023/any-metric-receiver/#fine-tuning-with-the-transform-processor

```
processors:
  transform:
    metric_statements:
      - context: datapoint
        statements:
          - set(metric.unit, attributes["unit"])
          - delete_key(attributes, "unit")
```

obs-gh-enricogiorio: Ok so this uses a lower level context. This will be more expensive for sure, I'll try it out. Thanks!

obs-gh-enricogiorio: I will open an issue regarding the `metric` context btw, because it explicitly says that you can access the unit, but it clearly doesn't work.

obs-gh-alexlew: I haven't tried it, just found it in the opentelemetry docs

obs-gh-mattcotter:
@obs-gh-enricogiorio should we close this old PR? If this is still needed, I can take it over if you can point me to a metric in O2 with missing units.

obs-gh-mattcotter:
closing old prs

PR #283, opened 2024-12-13
ci: update test logs for json error parsing

obs-gh-nikhildua:
Addresses https://observe.atlassian.net/browse/OB-38641

`test_logs.py` has been updated to parse logs as JSON (they are set as JSON in debug) and extract `level: <>` information for checks

### DIFF
--- a/integration/scripts/test_logs.py
+++ b/integration/scripts/test_logs.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import pytest
+import warnings
 
 @pytest.mark.tags(
         "default.yaml",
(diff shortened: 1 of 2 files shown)


integration/scripts/test_logs.py
```diff
@@ -11,23 +12,10 @@ def test_errors_logs(kube_client, helm_config):
...
 
     This test does the following:
-    - Check that all pods in the cluster's namespace have no errors/warnings in their logs
-    - Ignore expected errors
+    - Check that all pods in the cluster's namespace have no 'level': 'error' in their json logs
     - Check that all pods have the expected logs
     """
 
-
-    ignore_error_patterns = [
```
obs-gh-alexlew, line 20: were all these old messages in the warning level? or they were error level and they'd actually fail a test now?

obs-gh-nikhildua: These we old ignore patterns we had to manually exclude these warnings (mainly level:warn)

obs-gh-alexlew: if any of them were level:error then they would now start failing the integration tests since we fail any level:error now right? is that what you want?

obs-gh-alexlew (approved):

integration/scripts/test_logs.py
obs-gh-nikhildua: Correct. These would now fail 
The patterns added were mainly warnings I believe (which had the word "Error" in it)
Instead of checking for presence of "error" and failing that, we just check for `level'

I can update accordingly if we see more failures but the end goal is that `level: error` should be a red flag for us to "do something" vs `level: warn` which we were previously failing on but manually ignoring

PR #289, opened 2024-12-18
feat: allow custom otel config to be injected to services

obs-gh-alexlew:
This change allows for additional OTEL Config to be provided for each of the 4 deployments of the Observe Agent in the Helm chart. The custom config will be take precedence and will be merged with the config generated from the chart templates for each of the services.

### DIFF
--- a/charts/agent/templates/_config.tpl
+++ b/charts/agent/templates/_config.tpl
@@ -1,0 +1,27 @@
+{{- define "observe.deployment.applyClusterEventsConfig" -}}
+{{- $values := deepCopy .Values }}
+{{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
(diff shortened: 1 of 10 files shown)


charts/agent/templates/_config.tpl
```diff
@@ -0,0 +1,27 @@
+{{- define "observe.deployment.applyClusterEventsConfig" -}}
```
arthur-observe, line 1: Need example

charts/agent/templates/_config.tpl
```diff
@@ -0,0 +1,27 @@
...
+{{- $values := deepCopy .Values }}
+{{- $data := dict "Values" $values | mustMergeOverwrite (deepCopy .) }}
```
obs-gh-alexlew, line 3: I was misunderstanding how pipelines `|` work, this ends up taking the new dict `{ Values: (ouroldvalues) }` and using it as the last argument to mustMergeOverwrite (so the source that takes precedence). Think this is all in the name of using deepcopies for everything so that we dont accidentally change the global Values

arthur-observe (approved):

PR #292, opened 2024-12-18
fix: cleanup some duplicated configs

obs-gh-alexlew:


charts/agent/templates/_cluster-events-config.tpl
```diff
@@ -123,9 +123,8 @@ processors:
           - set(attributes["observe_transform"]["facets"]["cronJobName"], attributes["observe_transform"]["facets"]["jobName"]) where IsMatch(attributes["observe_transform"]["facets"]["jobName"], ".*-\\d{8}$")
           - replace_pattern(attributes["observe_transform"]["facets"]["cronJobName"], "-\\d{8}$$", "")
           - set(attributes["observe_transform"]["facets"]["statefulSetName"], body["metadata"]["ownerReferences"][0]["name"]) where body["metadata"]["ownerReferences"][0]["kind"] == "StatefulSet"
-          - set(attributes["observe_transform"]["facets"]["deploymentName"], body["metadata"]["ownerReferences"][0]["name"]) where body["metadata"]["ownerReferences"][0]["kind"] == "ReplicaSet"
```
obs-gh-enricogiorio, line 126: I remember there might have been a reason for this. Could it be because if the owner is a ReplicaSets, then it's a different name pattern that wouldn't match the regex? Please double check.

charts/agent/templates/_config-processors.tpl
```diff
@@ -22,7 +22,6 @@ k8sattributes:
     - k8s.daemonset.name
     - k8s.cronjob.name
     - k8s.job.name
-    - k8s.node.name
```
obs-gh-enricogiorio, line 25: Why remove it? Is it because it's not needed?

obs-gh-mattcotter:
closing old prs

PR #296, opened 2025-01-07
feat: add multiline to filelog config

obs-gh-alexlew:


arthur-observe (commented):
Can we add one working multiline example to repo with change?

arthur-observe (approved):

PR #299, opened 2025-01-15
chore: Add example for containerd multiline timestamp and java CS-1402

obs-gh-justindaines:
Added example using the default Java parser.

Added example matching timestamps with any of the following timestamps: 2024-12-03 07:31:12,563
07:31:20,550
[2024-12-03 07:31:28.353]


obs-gh-nikhildua:
Would be great to show a few screenshots of end results in the example!

obs-gh-felipequiros (approved):
Changes look good.

PR #300, opened 2025-01-15
feat: add forwarding daemonset and trace support

obs-gh-alexlew:


charts/agent/values.yaml
```diff
@@ -491,6 +498,18 @@ node-logs-metrics:
   tolerations: []
 
   ports:
+    otlp:
```
obs-gh-alexlew, line 501: need to disable these to free up the host ports on each node. for some reason only the existing daemonset has these mapped as host ports, the rest do not even though we're not explicitly disabling them. https://github.com/open-telemetry/opentelemetry-helm-charts/blob/dbddf8420986a72c0cb389edbce4d8b151841937/charts/opentelemetry-collector/values.yaml#L269 the defaults set a hostport but maybe it doesn't apply to non-daemonsets?

arthur-observe (approved):

PR #303, opened 2025-01-16
fix: upsert uidOverride if provided

obs-gh-alexlew:


obs-gh-alexlew:
tested with nil value 

```
cluster:
  uidOverride:
    value:
```

and fallback to clusterUid from configmap still works.

xueweiz (approved):
Thanks!

PR #307, opened 2025-01-23
fix: use more general pattern for exclude filelog default

obs-gh-alexlew:


obs-gh-alexlew:
for some reason the existing config is not causing tmp files to be excluded, even with paths that should match i.e. `"/var/log/pods/prod_transformer-0_0c7bbf73-ae27-4c9c-a689-061e2d4edaf7/transformer/0.log.20250122-232849.tmp"` 

i think maybe the include part is actually changing what the path which then gets passed to the exclude looks like. this more general path should hopefully catch these files

obs-gh-alexlew:
this also matches the default config in vector https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#exclude_paths_glob_patterns

arthur-observe (approved):

PR #308, opened 2025-01-23
feat: forward metrics as prometheus metrics instead

obs-gh-alexlew:


charts/agent/templates/_config-exporters.tpl
```diff
@@ -48,7 +48,7 @@ otlphttp/observe/forward/trace:
 {{- end -}}
 
 {{- define "config.exporters.prometheusremotewrite" -}}
-prometheusremotewrite:
+prometheusremotewrite/observe:
```
obs-gh-alexlew, line 51: renaming this and adding a suffix to lower probability of overlap in overrides

charts/agent/templates/_forwarder-config.tpl
```diff
@@ -60,11 +63,11 @@ service:
     logs/observe-forward:
       receivers: [otlp/app-telemetry]
       processors: [memory_limiter, k8sattributes, batch, resourcedetection/cloud, resource/observe_common, attributes/debug_source_app_logs]
-      exporters: [{{ join ", " $logsMetricsExporters }}]
+      exporters: [{{ join ", " $logsExporters }}]
     metrics/observe-forward:
       receivers: [otlp/app-telemetry]
       processors: [memory_limiter, k8sattributes, batch, resourcedetection/cloud, resource/observe_common, attributes/debug_source_app_metrics]
```
obs-gh-mattcotter, line 69: It's safer in this case to add the `deltatocumulative` processor here in case users forward metrics with delta temporality

obs-gh-mattcotter (approved):
One suggestion, otherwise lgtm

obs-gh-mattcotter (approved):

PR #312, opened 2025-02-03
fix: Update LUA and multiline

obs-gh-justindaines:
Update LUA and multiline to handle `stdout F` and `stdout P`


obs-gh-felipequiros (approved):
Looks good. Change captures stdout P  and F.

PR #313, opened 2025-02-03
feat: enable forwarder daemonset by default

obs-gh-alexlew:


integration/modules/deploy_helm/helm.tf
```diff
@@ -13,6 +13,7 @@ resource "helm_release" "observe-agent" {
       {
         observe_url                     = var.observe_url,
         observe_token                   = var.observe_token,
+        trace_token                     = var.observe_token,
```
obs-gh-nikhildua, line 16: If trace token is now becoming part of default helm deployment, it needs to be defined in this file 
https://github.com/observeinc/helm-charts/blob/main/integration/modules/deploy_helm/variables.tf 

Not sure why its specifically showing the "observe" namespace error but my guess is that it has to do something with this variable (or default) not being defined. Terraform plan is probably borked and the tftests don't really do plan/apply verbose like regular terraform

obs-gh-mattcotter (commented):
lgtm other than @obs-gh-nikhildua 's comment

obs-gh-mattcotter (approved):

PR #318, opened 2025-02-05
feat(multiline): add multiline examples

obs-gh-johnhwang:
CS-1844
add multiline examples


arthur-observe (changes requested):
I would expect an example to include a sample log generator deployment and a sample agent config - am I missing something?

obs-gh-johnhwang:
That’s my fault. I interpreted it wrong. I will get it corrected.

obs-gh-abhishekrao (approved):

arthur-observe (approved):

PR #323, opened 2025-02-12
feat: lower poll interval on filelog

obs-gh-alexlew:


charts/agent/templates/_node-logs-metrics-config.tpl
```diff
@@ -124,6 +124,7 @@ receivers:
   {{ end -}}
   {{- if .Values.node.containers.logs.enabled }}
   filelog:
+    poll_interval: 5ms
```
obs-gh-mattcotter, line 127: Is this too frequent? I don't know how expensive it is to check the files... [the default is 200ms](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md#configuration), maybe we could do 50ms for 4x frequency instead of 40x? If you tested this and aren't concerned though, then lgtm

obs-gh-mattcotter (dismissed):

charts/agent/templates/_node-logs-metrics-config.tpl
obs-gh-alexlew: gonna hold off on merging this until after we test it out internally

obs-gh-mattcotter (approved):

PR #327, opened 2025-03-17
feat: add benchmark test and accompanying utils

obs-gh-mattcotter:
This adds a benchmark script which can be run locally to generate agent benchmark data. It has four load generators, each of which can be tuned to various volume levels. Example run:
```
$ HELM_NAMESPACE=observe-test pytest ./integration/scripts/test_benchmark.py
```


obs-gh-mattcotter:
> looks good but with all these tools being added I'd love to see a README.md that covers how it works/how to use it in the future

Added! https://github.com/observeinc/helm-charts/blob/617de28ccc7e92bdfacd69b2d98a20cf29469806/integration/scripts/benchmark/README.md

integration/scripts/benchmark/README.md
```diff
@@ -0,0 +1,35 @@
+# `observe-agent` helm chart benchmarking
+
+## Usage
+
+This package is designed to benchmark the `observe-agent` helm chart. To run the benchmarking tests, you will need to:
+
+1. Create / connect to a kubernetes cluster
+2. Install the `observe-agent` helm chart
+3. Run the tests via: `HELM_NAMESPACE=observe-test pytest ./integration/scripts/test_benchmark.py`
```
obs-gh-alexlew, line 9: how do I provide the credentials for my specific kubernetes cluster?

obs-gh-mattcotter: Good point! I added:
> The benchmarking script will automatically connect to the current kubernetes context. View the current context via `kubectl config current-context`, set this via `kubectl config use-context <context>`.

Btw I'm force pushing because the build fails if the PR has multiple commits :( Can we change that and enforce that the commits are squashed when landed to main instead?

obs-gh-alexlew: i prefer squashing when landing as well, i didn't set up this repo and not sure how to remove that check + add squashing

obs-gh-alexlew (approved):

PR #329, opened 2025-03-18
fix: tune default resource values after benchmark testing

obs-gh-mattcotter:
This is inline with the usage we observed in our 10k DPS high volume scenario.

Related to: https://github.com/observeinc/helm-charts/pull/327

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -163,17 +163,17 @@ cluster-events:
   configMap:
     create: false
     existingName: "cluster-events"
+
+  resources:
+    requests:
+      cpu: 100m
+      memory: 128Mi
+    limits:
+      memory: 128Mi
   # ----------------------------------------- #
 
   # ----------------------------------------- #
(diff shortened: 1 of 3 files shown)


charts/agent/values.yaml
```diff
@@ -435,7 +435,7 @@ node-logs-metrics:
       cpu: 250m
       memory: 256Mi
     limits:
-      memory: 256Mi
```
obs-gh-alexlew, line 438: I forgot this earlier, we actually I think prefer to set memory limits == memory requests. https://home.robusta.dev/blog/kubernetes-memory-limit see for more detail. But let's pick the higher number and use that as request and limits.

obs-gh-alexlew:
also you'll need to update the chart version in `chart.yaml` and then run `pre-commit run --show-diff-on-failure --color=always --all-files` to generate the readme and update your commit

obs-gh-alexlew (approved):

PR #330, opened 2025-03-19
feat: remove some resource labels from prometheus metrics

obs-gh-alexlew:


charts/agent/templates/_config-processors.tpl
```diff
@@ -32,17 +32,32 @@ k8sattributes:
     - k8s.pod.uid
     - k8s.cluster.uid
     - k8s.node.name
-    - k8s.node.uid
     - k8s.container.name
     - container.id
+    labels:
+    - tag_name: app.kubernetes.io/name
```
xueweiz, line 38: WOW AWESOME, thanks for doing this!

xueweiz (approved):

charts/agent/templates/_cluster-metrics-config.tpl
```diff
@@ -150,6 +136,28 @@ processors:
 {{- include "config.processors.attributes.k8sattributes" . | nindent 2 }}
 
 {{- include "config.processors.resource.observe_common" . | nindent 2 }}
+  resource/drop_additional_pod_metrics_labels:
```
obs-gh-mattcotter, line 139: Are we reasonably confident that this won't break any downstream datasets customers could have created?

obs-gh-mattcotter (approved):

charts/agent/templates/_cluster-metrics-config.tpl
obs-gh-alexlew: There's no way to know really. These labels existed and after this they won't so if they are relying/filtering on them then that would break things. We could also draft this change as a major version change (and ideally roll it together with the removal of `TRACE_TOKEN`)

PR #332, opened 2025-03-24
feat: separate memory-limiters for custom pod metrics and cluster met…

obs-gh-alexlew:
…rics

### DIFF
--- a/charts/agent/templates/_cluster-metrics-config.tpl
+++ b/charts/agent/templates/_cluster-metrics-config.tpl
@@ -143,13 +143,21 @@ receivers:
   {{ end }}
 
 processors:
-{{- include "config.processors.memory_limiter" . | nindent 2 }}
 
 {{- include "config.processors.batch" . | nindent 2 }}
 
 {{- include "config.processors.attributes.k8sattributes" . | nindent 2 }}
 
 {{- include "config.processors.resource.observe_common" . | nindent 2 }}
+  memory_limiter/pod_metrics:
+    check_interval: 5s
+    limit_percentage: 60
+    spike_limit_percentage: 5
+
+  memory_limiter/cluster_metrics:
+    check_interval: 5s
+    limit_percentage: 30
+    spike_limit_percentage: 5
 
   # attributes to append to objects
   attributes/debug_source_cluster_metrics:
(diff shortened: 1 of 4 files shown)


obs-gh-mattcotter (dismissed):

charts/agent/values.yaml
```diff
@@ -299,9 +299,9 @@ cluster-metrics:
   resources:
     requests:
       cpu: 250m
-      memory: 256Mi
+      memory: 512Mi
```
obs-gh-alexlew, line 302: need to bump this since we want at least 128mb of ram specifically for cluster_metrics and that's going to be limited to 30% (and usually 25 pre-spike) of total so 0.25 * 512 ~128. Old value would have had it at 64 which is lower than the high volume examples.

charts/agent/templates/_cluster-metrics-config.tpl
```diff
@@ -143,13 +143,21 @@ receivers:
...
 
 {{- include "config.processors.resource.observe_common" . | nindent 2 }}
+  memory_limiter/pod_metrics:
+    check_interval: 5s
+    limit_percentage: 60
+    spike_limit_percentage: 5
+
+  memory_limiter/cluster_metrics:
+    check_interval: 5s
+    limit_percentage: 30
```
obs-gh-alexlew, line 159: These two separate percentages add up to 90 now

obs-gh-alexlew:
Issue we're trying to resolve is that when pod metrics fill up the memory limiter we start losing cluster metrics as well. We want cluster metrics to be sent and limited independently of the prom pod metric scraping pipeline

charts/agent/values.yaml
obs-gh-mattcotter: Good catch! Are we concerned about other pods having pipelines with memory limiters that total more than 100%?

obs-gh-mattcotter (dismissed):

charts/agent/values.yaml
obs-gh-alexlew: This is the only pod that has more than one memory limiter, rest should be configured naturally below 100

charts/agent/templates/_cluster-metrics-config.tpl
obs-gh-mattcotter: Oh okay. And we haven't had any concerns about limiting separately elsewhere?

obs-gh-alexlew: yup this is the one spot so far that its come up since the pod metrics actually fills up. if we need to split it elsewhere we can look at that separately

obs-gh-mattcotter (approved):

PR #333, opened 2025-03-25
feat: add forwarder override config to values.yaml

obs-gh-alexlew:


charts/agent/values.yaml
```diff
@@ -144,6 +144,8 @@ agent:
     nodeLogsMetrics:
     # additional otel collector config for monitor deployment
     monitor:
+    # additional otel collector config for forwarder daemonset
+    forwarder:
```
obs-gh-mattcotter, line 148: Just to double check, this is already referenced here: https://github.com/observeinc/helm-charts/pull/300/files#diff-67c001953328ebda3ac9fc9284bab2d91e5f34c2244c3645b368ba62b2fd106dR25
so it will work without any other changes right?

obs-gh-mattcotter (approved):

charts/agent/values.yaml
obs-gh-alexlew: yup exactly I just forgot to add the top level values.yaml field

PR #337, opened 2025-03-31
fix: Replace kubeletstats to use node IP address - (OBSSD-482)

obs-gh-justindaines:
This PR addresses an issue where the `kubeletstats` receiver in the agent fails to scrape metrics, resulting in a `connection refused` error. The failure occurs because the current configuration uses the node's hostname (via `K8S_NODE_NAME`), which resolves to the loopback address (`127.0.1.1`) instead of the node's actual IP. This prevents the receiver from reaching the kubelet’s `/stats/summary` endpoint on port 10250.

Added an override in for `node.kubeletstats.useNodeIp: true` to update the endpoint to use `K8S_NODE_IP` if set or default to `K8S_NODE_NAME`.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.50.0
+version: 0.50.1
 appVersion: "2.2.1"
 dependencies:
   - name: opentelemetry-collector
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.50.0](https://img.shields.io/badge/Version-0.50.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.2.1](https://img.shields.io/badge/AppVersion-2.2.1-informational?style=flat-square)
+![Version: 0.50.1](https://img.shields.io/badge/Version-0.50.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.2.1](https://img.shields.io/badge/AppVersion-2.2.1-informational?style=flat-square)
 
 Chart to install K8s collection stack based on Observe Agent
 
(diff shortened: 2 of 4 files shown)


obs-gh-alexlew:
is the impact that kubeletstats doesn't work at all right now?

obs-gh-mattcotter (dismissed):
Nice find! lgtm

obs-gh-justindaines:
> > is the impact that kubeletstats doesn't work at all right now?
> 
> Correct, the `kubeletstats` are not being collected due to DNS resolving to the loopback.
> 
> > 2025-03-28T16:27:07.133Z error kubeletstatsreceiver/scraper.go:103 call to /stats/summary endpoint failed {"kind": "receiver", "name": "kubeletstats", "data_type": "metrics", "error": "Get "https://<K8S_NODE_NAME>:10250/stats/summary": dial tcp 127.0.1.1:10250: connect: connection refused"}
> 
> Similar solution - [open-telemetry/opentelemetry-collector-contrib#26481 (comment)](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26481#issuecomment-1720797914)
> 
> Let me try using the suggestion - [open-telemetry/opentelemetry-collector-contrib#26481 (comment)](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/26481#issuecomment-1721421682) instead.

After some additional consideration, I am hesitant to update 

> open-telemetry/opentelemetry-collector-contrib#26481 (comment)

After further consideration and review, I am hesitant to drop the default since `open-telemetry/opentelemetry-collector-contrib` is using in [endpoint: '${env:K8S_NODE_NAME}:10250'](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b1ea2d22d262bb98eb9eb635bab088300c136612/examples/kubernetes/daemonset-collector-dev.yaml#L42) 

I am planning on adding an override in for `node.kubeletstats.useNodeIp: true` to update the endpoint to use `K8S_NODE_IP` if set or default to `K8S_NODE_NAME`. Since the `connection refused` scenario does not appear to occur frequently.

Example:
```
endpoint: {{ if .Values.node.kubeletstats.useNodeIp }}"${env:K8S_NODE_IP}:10250"{{ else }}"${env:K8S_NODE_NAME}"{{ end }}

Validating my new changes now.

obs-gh-alexlew (dismissed):

obs-gh-alexlew:
change seems a lot safer this way. when specifically does `K8S_NODE_NAME` not work?

obs-gh-justindaines:
> change seems a lot safer this way. when specifically does `K8S_NODE_NAME` not work?

In my scenario, CoreDNS is utilizing the `/etc/hosts` file and resolving the `K8S_NODE_NAME` to the IP `127.0.1.1`. DNS is mapping the hostname to this loopback address, which then rejects the connection. The connection should target the Node IP, not the Pod itself.

PR #343, opened 2025-04-04
fix: Update Makefile to include `agent`

obs-gh-justindaines:
Update Makefile to include `agent`.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.51.2
-appVersion: "2.2.1"
+version: 0.51.3
+appVersion: "2.2.2"
 dependencies:
   - name: opentelemetry-collector
     version: 0.101.1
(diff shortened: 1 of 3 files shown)


charts/agent/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.51.1
-appVersion: "2.2.1"
+version: 0.51.2
+appVersion: "2.2.2"
```
obs-gh-mattcotter, line 6: Why is the agent version changing? Is this just to appease some integration test?

obs-gh-justindaines: Exactly. For some reason, `helm upgrade` didn’t fetch the updated chart. I incremented the `appVersion` to ensure Helm recognizes the changes and rules out any issues of the `helm upgrade` not detecting any meaningful updates.

obs-gh-mattcotter: But only the makefile is changing right? Why would you need a helm upgrade after this?

obs-gh-justindaines:
> @obs-gh-justindaines sorry for the slow reply. I think we should avoid bumping the chart version when the chart itself isn't changing. I'm still a bit confused by the helm issue you mentioned; is this tracked in a ticket somewhere?
> 
> I copied your changes into #345, in which I also bumped the chart minor version. Will that allow you to keep debugging and close this?

Perfect, I'll close this one then. Thank you!

PR #346, opened 2025-04-14
fix: remove separate memorylimiters

obs-gh-alexlew:


obs-gh-alexlew:
turns out memory limiters are process level no matter what so this actually just breaks things

PR #348, opened 2025-04-15
feat: rely on observe-agent config for health check and internal telemetry

obs-gh-mattcotter:
OB-43962
Putting these values in the agent config makes `observe-agent status` work by default.


obs-gh-mattcotter:
@obs-gh-alexlew one consideration I want to call out is that this change makes our helm setup more reliant on the `observe-agent`. We can always work around this by adding health check and internal telemetry config back into a custom values.yaml file, but it adds more steps to byo collector. I think a better common experience is worth it, but lmk if you have concerns.

PR #354, opened 2025-04-22
fix: lower memory-limiter to 80

obs-gh-alexlew:
GOMEMLIMIT gets automatically set to 80% of the Kube resources so this should be 80% total as well

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -81,22 +81,9 @@ resource/observe_common:
 {{- define "config.processors.memory_limiter" -}}
 # https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/memorylimiterprocessor/README.md
 memory_limiter:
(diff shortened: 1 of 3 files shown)


charts/agent/templates/_config-processors.tpl
```diff
@@ -91,7 +91,7 @@ memory_limiter:
   # This option is used to calculate memory_limit from the total available memory.
   # For instance setting of 75% with the total memory of 1GiB will result in the limit of 750 MiB.
   # The fixed memory setting (limit_mib) takes precedence over the percentage configuration.
-  limit_percentage: 90
+  limit_percentage: 80
```
obs-gh-mattcotter, line 94: > GOMEMLIMIT gets automatically set to 80% of the Kube resources so this should be 80% total as well

^ Can you add the PR description as a comment here?

obs-gh-mattcotter (dismissed):

obs-gh-mattcotter (approved):

PR #363, opened 2025-05-12
feat: add parallelism configs and feature gate for prwe

obs-gh-alexlew:


charts/agent/values.yaml
```diff
@@ -321,7 +321,7 @@ cluster-metrics:
 
   command:
     name: "observe-agent"
-    extraArgs: ["start", "--observe-config=/observe-agent-conf/observe-agent.yaml", "--config=/conf/relay.yaml"]
+    extraArgs: ["start", "--observe-config=/observe-agent-conf/observe-agent.yaml", "--config=/conf/relay.yaml", "--feature-gates=+exporter.prometheusremotewritexporter.EnableMultipleWorkers"]
```
obs-gh-alexlew, line 324: these could theoretically be removed once the agent change goes in and we bump agent version in helm chart but im wondering if its positive to be explicit here or its just redundant

obs-gh-mattcotter: I think for a few versions it's better to be explicit, but by the time the feature gate gets promoted, we can remove this if we haven't yet.

obs-gh-mattcotter (approved):

PR #368, opened 2025-05-14
fix: default to prom scraping in cluster-metrics

obs-gh-alexlew:


charts/agent/Chart.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
     version: 0.101.1
     repository: https://open-telemetry.github.io/opentelemetry-helm-charts
     alias: prometheus-scraper
-    condition: application.prometheusScrape.enabled
+    condition: application.prometheusScrape.independentDeployment
```
obs-gh-alexlew, line 22: kind of painful you can't AND conditions here, really it should be enabled: true and independentDeployment: true but can only have one field here. so if enabled is false we just set the exporter to `nop` instead but instantiate the service, cm everything

charts/agent/templates/_prometheus-scraper-config.tpl
```diff
@@ -146,10 +31,17 @@ processors:
...
-      processors: [memory_limiter, k8sattributes, batch, resource/observe_common, attributes/debug_source_pod_metrics]
-      exporters: [{{ join ", " $podMetricsExporters }}]
+{{- if and (eq .Values.application.prometheusScrape.enabled true) (eq .Values.application.prometheusScrape.independentDeployment true) }}
+      metrics/pod_metrics:
+        receivers: [prometheus/pod_metrics]
+        processors: [memory_limiter, k8sattributes, batch, resource/observe_common, attributes/debug_source_pod_metrics]
+        exporters: [{{ join ", " $podMetricsExporters }}]
+{{ else }}
+      metrics:
+        receivers: [prometheus/pod_metrics]
```
obs-gh-mattcotter, line 41: Better to use `nop` here too and skip processors? Is this pipeline doing anything?

obs-gh-mattcotter (approved):

obs-gh-mattcotter (approved):

PR #369, opened 2025-05-14
fix: update filestorage mount

obs-gh-alexlew:


obs-gh-alexlew:
So this one is weird, it seems to work right now anyways? and if I go into the container and ls i get 
```
/ # ls /var/lib
docker         misc           observe-agent  otelcol
```
but filestorage is only defined in `observe-agent`. 

It almost seems like this shouldn't actually work so I'm kind of confused why the current state is working. The agent also expects the path to be `/var/lib/observe-agent/filestorage`

obs-gh-mattcotter:
> So this one is weird, it seems to work right now anyways? and if I go into the container and ls i get
> 
> ```
> / # ls /var/lib
> docker         misc           observe-agent  otelcol
> ```
> 
> but filestorage is only defined in `observe-agent`.
> 
> It almost seems like this shouldn't actually work so I'm kind of confused why the current state is working. The agent also expects the path to be `/var/lib/observe-agent/filestorage`

How did you end up running into this? Instead of changing the mounts, can we override the dir in the extension configuration: https://github.com/observeinc/helm-charts/blob/a69f83796a219e3c1dde0940a5e294f12a63f43b/charts/agent/templates/_config-extensions.tpl#L1-L4

I think that would better preserve backwards compatibility.

obs-gh-alexlew:
just forwarded you the customer email

obs-gh-mattcotter:
I still think changing the bundled config instead of the mount is better for backwards compatibility.

PR #371, opened 2025-05-19
feat: add an option to create generic otel config overrides for all agent configmaps at once

obs-gh-mattcotter:


obs-gh-mattcotter:
I tested this by adding the following to a values.yaml file:
```yaml
agent:
  config:
    global:
      overrides:
        receivers:
          nop/test:
```

and then running `helm template agent ./charts/agent -f ./nop_values.yaml` and checking the diff of the rendered template before/after this change; the new receiver was added to all six configmaps and nothing else changed.

obs-gh-alexlew (approved):

PR #372, opened 2025-05-20
feat: add flag to modify forwarder metrics output format

obs-gh-alexlew:


charts/agent/templates/_forwarder-config.tpl
```diff
@@ -47,14 +48,28 @@ processors:
...
 {{- $traceExporters := (list "otlphttp/observe/forward/trace") -}}
 {{- $logsExporters := (list "otlphttp/observe/base") -}}
-{{- $metricsExporters := (list "prometheusremotewrite/observe") -}}
+{{- $metricsExporters := (list) -}}
+
+{{ if eq .Values.node.forwarder.metrics.outputFormat "prometheus" -}}
+  {{- $metricsExporters = concat $metricsExporters ( list "prometheusremotewrite/observe" ) | uniq }}
+{{- else if eq .Values.node.forwarder.metrics.outputFormat "otel" -}}
+  {{- $metricsExporters = concat $metricsExporters ( list "otlphttp/observe/otel_metrics" ) | uniq }}
+{{- end }}
```
obs-gh-mattcotter, line 57: Can we add an `else` case here to error if an invalid value is provided? I think [fail](https://helm.sh/docs/chart_template_guide/function_list/#fail) might be able to do that.

PR #374, opened 2025-05-20
internal: rm single commit check

obs-gh-alexlew:


obs-gh-alexlew:
gonna just remove this check wholesale, its wrapped in our own custom action thats pretty specific for single commits

obs-gh-mattcotter (approved):

PR #381, opened 2025-06-13
feat: add cadvsior prometheus scraping config

obs-gh-alexlew:


charts/agent/templates/_node-logs-metrics-config.tpl
```diff
@@ -157,6 +157,28 @@ receivers:
...
+          scheme: https
+          tls_config:
+            ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+            insecure_skip_verify: true
+          bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
+
+          kubernetes_sd_configs:
+            - role: node
+
+          relabel_configs:
```
obs-gh-jabernet, line 174: what do those two configs do? did we copy them from some previous setup? we had some problematic configs recently with the prom histograms, just want to make sure we don't run into some similar issues again

charts/agent/templates/_node-logs-metrics-config.tpl
```diff
@@ -215,6 +244,12 @@ service:
         processors: [memory_limiter, k8sattributes, batch, resourcedetection/cloud, resource/observe_common, attributes/debug_source_kubeletstats_metrics]
         exporters: [{{ join ", " $kubeletstatsExporters }}]
       {{- end -}}
+      {{- if .Values.node.metrics.cadvisor.enabled }}
+      metrics/cadvisor:
+        receivers: [prometheus/cadvisor]
+        processors: [memory_limiter, k8sattributes, batch, resourcedetection/cloud, resource/observe_common, attributes/debug_source_cadvisor_metrics]
```
obs-gh-jabernet, line 250: can be done as a follow up but i wonder if it would make sense here to use https://opentelemetry.io/docs/collector/building/connector/ to batch all the 3 metric types together and only have the `attributes/debug_source_xyz` processor different? since this is a daemonset it would make sense for both customer and our infra to reduce http calls

obs-gh-jabernet (commented):
thanks for quickly implementing this! looks good in general, just wanted to make sore the relabeling configs are critical before approving

charts/agent/templates/_node-logs-metrics-config.tpl
obs-gh-alexlew: basically the first part just says scrape only that address and the second part says only target the cadvisor endpoint

obs-gh-alexlew: not sure if there's a canonical example in the docs but here's an online source that's similar https://www.cloudforecast.io/blog/cadvisor-and-kubernetes-monitoring-guide/

charts/agent/templates/_node-logs-metrics-config.tpl
obs-gh-alexlew: Yup it would make sense, would need to refactor the logic that builds the exporter array based on the flags, initially it was simpler to just have the flag control the inclusion/exclusion of the pipeline instead. will file a followup ticket

obs-gh-jabernet (approved):

PR #383, opened 2025-06-18
feat: use otel semantic extraction for service attributes

obs-gh-mattcotter:


obs-gh-alexlew:
did you test this and the attributes we had before all still show up?

obs-gh-alexlew (approved):

obs-gh-mattcotter:
> did you test this and the attributes we had before all still show up?

Yes! Most of this is additive; the `annotations` regex that I'm removing was copied verbatim from the `otel_annotations` feature before it was released. The only label we're losing is `service.name`, which I verified is still generated locally with the new metadata setting.

PR #389, opened 2025-06-30
fix: add hash calculation

icco:
This sets the existingPath to go with existingName so that hash calculation happens.

See the following code for reference:

- https://github.com/open-telemetry/opentelemetry-helm-charts/blob/a3074b4a0c9839054488ff08d5daabeb5d89d522/charts/opentelemetry-collector/values.yaml#L87-L89 
- https://github.com/open-telemetry/opentelemetry-helm-charts/blob/a3074b4a0c9839054488ff08d5daabeb5d89d522/charts/opentelemetry-collector/templates/_helpers.tpl#L234-L249

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -188,6 +188,7 @@ cluster-events:
   configMap:
     create: false
     existingName: "cluster-events"
+    existingPath: "cluster-events-configmap.yaml"
 
   resources:
     requests:
@@ -320,6 +321,7 @@ cluster-metrics:
   configMap:
     create: false
     existingName: "cluster-metrics"
+    existingPath: "cluster-metrics-configmap.yaml"
 
   resources:
     requests:
@@ -451,6 +453,7 @@ prometheus-scraper:
   configMap:
     create: false
     existingName: "prometheus-scraper"
+    existingPath: "prometheus-scraper-configmap.yaml"
 
   resources:
     requests:
@@ -582,6 +585,7 @@ node-logs-metrics:
   configMap:
     create: false
     existingName: "node-logs-metrics"
+    existingPath: "node-logs-metrics-configmap.yaml"
 
   resources:
     requests:
@@ -764,6 +768,7 @@ monitor:
   configMap:
     create: false
     existingName: "monitor"
+    existingPath: "agent-monitor-configmap.yaml"
 
   resources:
     requests:
@@ -894,6 +899,7 @@ forwarder:
   configMap:
     create: false
     existingName: "forwarder"
+    existingPath: "forwarder-configmap.yaml"
 
   resources:
     requests:


obs-gh-mattcotter:
@icco thanks for your pull request! Were you able to get this to work locally? When I try it, the subchart is unable to find the template since it looks for a relative path. Here's the error I see:
```
Error: INSTALLATION FAILED: template: agent/charts/prometheus-scraper/templates/deployment.yaml:35:12: executing "agent/charts/prometheus-scraper/templates/deployment.yaml" at <include "opentelemetry-collector.configTemplateChecksumAnnotation" .>: error calling include: template: agent/charts/cluster-events/templates/_helpers.tpl:239:22: executing "opentelemetry-collector.configTemplateChecksumAnnotation" at <include (print $.Template.BasePath "/" .Values.configMap.existingPath) .>: error calling include: template: no template "agent/charts/prometheus-scraper/templates/prometheus-scraper-configmap.yaml" associated with template "gotpl"
```

Seems like someone else has had this same issue when using the collector chart as a subchart:
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1302#discussion_r2168372227
If you have this working, let's figure out what is missing in my setup and we can get this feature enabled.

obs-gh-mattcotter:
> I didn't have a local test environment for this so I didn't test it, but you're right, running helm template, I also see this error.
> 
> [open-telemetry/opentelemetry-helm-charts#1737](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1737) also highlights this issue. I tried reading through https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#sharing-templates-with-subcharts and I'm not enough of a helm expert to understand how to take advantage of this. I wonder if we could make a fix to the otel helm chart to make this work.
> 
> Do you have a test env setup to test changes to the otel chart with your observe chart?
> 
> Or should we try the hash option that the issue I linked suggests as a work around?

@icco I looked into the possibility of sharing templates, and my most promising lead was that we could re-define `opentelemetry-collector.configName` and change the name of the template referenced in the pod volumes. However, I hit a similar problem of not being able to pass variables into a template definition. For example, something like this would be great:
```yaml
{{ $forwarderConfig := include "observe.daemonset.applyForwarderConfig" . }}

{{- define "opentelemetry-collector.configName" -}}
    {{- if eq .Chart.Name "forwarder" -}}
        forwarder-{{ print $forwarderConfig | sha256sum }}
    {{- else -}}
        {{ .Chart.Name }}
    {{- end -}}
{{- end }}
```
but the result is `Error: parse error at (agent/templates/_config.tpl:47): undefined variable "$forwarderConfig"`.

The workaround of always restarting by adding a random annotation seems solid to me if that's the behavior that you want. You can do that with our current helm chart by adding this to your `values.yaml` file (or using it as a separate file):
```yaml
cluster-events:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"

prometheus-scraper:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"

cluster-metrics:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"

node-logs-metrics:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"

monitor:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"

forwarder:
  podAnnotations:
    randAlphaNum/autoRestartEverytime: "{{ randAlphaNum 5 }}"
```

I also was able to find a workaround using config map hashing when deploying with Kustomize, but that's a completely separate tool and wouldn't help unless you happen to be using it.

PR #391, opened 2025-07-03
test: add agent chart to helm install test

obs-gh-mattcotter:


test/test.sh
```diff
@@ -66,7 +66,7 @@ for chart in "$@"; do
     echo charts/$chart tests PASSED
     echo
     echo results:
-    $kc logs test-$chart
```
obs-gh-jabernet, line 69: this will ignore the error right? why? will we catch it later?

test/cmd/go.sum
obs-gh-jabernet, line 1: is this autogenerated or where are the dependencies coming from?

test/test.sh
obs-gh-mattcotter: We don't have a resource named `test-agent` in the agent chart, so there's nothing to get logs from. I don't think this was ever testing anything on purpose because it just calls `kubectl logs`.

PR #392, opened 2025-07-07
fix: Remove container information from pod metrics

obs-gh-jabernet:
Container information currently is not reliable, a random container id / name is being discovered, which can lead to confusion at best and labels/attributes dimensionality explosion at worst.

It should be possible to uniquely identify the container based on the pod, but I couldn't get this to work. That would still be the ideal solution - we can always add this back if we manage to get that to work. For now, better to remove the information completely.

Fixes OB-45911

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -34,9 +34,11 @@ k8sattributes:
       - k8s.pod.name
       - k8s.pod.uid
       - k8s.cluster.uid
+      {{- if (ne .target "pod_metrics") }}
       - k8s.container.name
-      {{- if ne .target "cluster_metrics" }}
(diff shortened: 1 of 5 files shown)


obs-gh-mattcotter (approved):

charts/agent/templates/_config-processors.tpl
```diff
@@ -34,10 +34,12 @@ k8sattributes:
       - k8s.pod.name
       - k8s.pod.uid
       - k8s.cluster.uid
+      {{- if (ne .target "pod_metrics") }}
```
obs-gh-alexlew, line 37: this nested NE condition is a little tough to understand, can we add a comment here to explain what happens in the cases

1. Not `pod_metrics or cluster_metrics`
2. `pod_metrics`
3. `cluster_metrics`

PR #393, opened 2025-07-07
fix: populate service.name correctly on scraped prometheus metrics

obs-gh-mattcotter:
Currently the prometheus receiver is setting the prom scrape job name as `service.name`:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.128.0/receiver/prometheusreceiver/internal/prom_to_otlp.go#L53

We need to delete this so `service.name` is set properly by `k8sattributes` processor.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.64.2
+version: 0.64.3
 appVersion: "2.5.0"
 dependencies:
   - name: opentelemetry-collector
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -87,6 +87,10 @@ bump-version:
 		fi; \
 	done
 
+.PHONY: pre-commit
```
obs-gh-mattcotter, line 90: I find this command so easy to forget, so adding it here to save time finding it (it's buried in a reference to a github action, so normally I just wait for it to fail)

obs-gh-jabernet (changes requested):
misses the most critical prom scraper in `_prometheus-scraper-config.tpl`

obs-gh-alexlew (approved):

obs-gh-jabernet (dismissed):

charts/agent/values.yaml
```diff
@@ -114,6 +114,10 @@ application:
     independentDeployment: false
     # scrape interval
     interval: 60s
+    # Use the prometheus service name for metric labels instead of populating the service name from k8s pod labels/annotations.
+    # If the scraped prometheus endpoints' target_info metrics are not configured to include the service name, the prometheus scrape
+    # job name will be used as the service name, which won't match other metrics for the pod. Setting this to false fixes that case.
+    keepPrometheusServiceName: true
```
obs-gh-mattcotter, line 120: @obs-gh-jabernet I made this an option since the target info metric can theoretically handle this.

obs-gh-jabernet (approved):

PR #394, opened 2025-07-08
feat: add option to enable RED metrics generation from spans

obs-gh-mattcotter:
This adds an option to generate RED metrics from spans in the agent forwarder. These will be used to power our Service Explorer and eventually allow standardized trace sampling.

I ran the benchmarking test with the medium volume trace data and low for everything else. The quick results show about a ~10% difference in CPU usage and <2% increase in RAM usage.

### DIFF
--- a/charts/agent/templates/_forwarder-config.tpl
+++ b/charts/agent/templates/_forwarder-config.tpl
@@ -2,7 +2,7 @@
 
 {{- $spanmetricsResourceAttributes := (list "service.namespace" "service.version" "deployment.environment" "k8s.pod.name" "k8s.cluster.uid" "k8s.namespace.name") -}}
 
-{{- if .Values.node.forwarder.redMetrics.enabled }}
+{{- if .Values.application.REDMetrics.enabled }}
 connectors:
   spanmetrics:
     aggregation_temporality: AGGREGATION_TEMPORALITY_DELTA
@@ -48,7 +48,7 @@ processors:
 {{- fail "Invalid maxSpanDuration for forwarder red metrics, valid values are 'none' or a number with a valid time unit: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/README.md#duration" }}
 {{- end }}
 
-{{- if .Values.node.forwarder.redMetrics.enabled }}
+{{- if .Values.application.REDMetrics.enabled }}
   attributes/debug_source_span_metrics:
     actions:
       - action: insert
@@ -202,7 +202,7 @@ service:
       receivers: [otlp/app-telemetry]
       processors:  [{{ join ", " $metricsProcessors }}]
       exporters: [{{ join ", " $metricsExporters }}]
-    {{- if .Values.node.forwarder.redMetrics.enabled }}
+    {{- if .Values.application.REDMetrics.enabled }}
     traces/spanmetrics:
       receivers:
         - otlp/app-telemetry
--- a/charts/agent/templates/_forwarder-config.tpl
+++ b/charts/agent/templates/_forwarder-config.tpl
@@ -48,48 +48,61 @@ processors:
 {{- end }}
 
 {{- if .Values.node.forwarder.redMetrics.enabled }}
-  # This handles schema normalization as well as moving status to attributes so it can be a dimension in spanmetrics
-  transform/shape_spans_for_red_metrics:
-    error_mode: ignore
-    trace_statements:
-      - set(span.attributes["peer.db.name"], span.attributes["db.system.name"]) where span.attributes["peer.db.name"] == nil and span.attributes["db.system.name"] != nil
-      - set(span.attributes["peer.db.name"], span.attributes["db.system"]) where span.attributes["peer.db.name"] == nil and span.attributes["db.system"] != nil
-      - set(span.attributes["status.message"], span.status.message) where span.status.message != ""
-
-  # This removes service.name for generated RED metrics associated with peer systems.
-  transform/remove_service_name_for_peer_metrics:
-    error_mode: ignore
-    metric_statements:
-      - delete_key(datapoint.attributes, "service.name") where datapoint.attributes["peer.db.name"] != nil or datapoint.attributes["peer.messaging.system"] != nil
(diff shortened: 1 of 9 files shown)


charts/agent/templates/_forwarder-config.tpl
```diff
@@ -14,7 +29,54 @@ receivers:
...
+  filter/drop_span_kinds_other_than_server_and_consumer_and_peer_client:
+    error_mode: ignore
+    traces:
+      span:
+        - span.kind == SPAN_KIND_CLIENT and span.attributes["peer.messaging.system"] == nil and span.attributes["peer.db.name"] == nil
+        - span.kind == SPAN_KIND_UNSPECIFIED
+        - span.kind == SPAN_KIND_INTERNAL
+        - span.kind == SPAN_KIND_PRODUCER
+
+{{- if eq .Values.node.forwarder.redMetrics.maxSpanDuration "none" }}
```
obs-gh-mattcotter, line 66: This value is configurable in Service Explorer, so we need to add a tuning nob in the helm chart in order to stay consistent.

charts/agent/templates/_forwarder-config.tpl
```diff
@@ -14,7 +29,54 @@ receivers:
...
+{{- else if (regexMatch "^[0-9]+(ns|us|ms|s|m|h)$" .Values.node.forwarder.redMetrics.maxSpanDuration) }}
+  # This drops spans that are longer than the configured time (default 1hr) to match service explorer behavior.
```
obs-gh-mattcotter, line 68: This value is configurable in Service Explorer, so we need to add a tuning nob in the helm chart in order to stay consistent.

integration/scripts/test_benchmark.py
```diff
@@ -90,12 +90,15 @@ def test_benchmark(
         pod.metadata.name for pod in pods.items if "agent" in pod.metadata.name
     ]
     agent_pods = sorted(agent_pods)
-    # The num nodes is computed assuming `num_pods = 3 + 2*num_nodes`
-    if len(agent_pods) < 5 or len(agent_pods) % 2 == 0:
+    num_deployments = 3
```
obs-gh-mattcotter, line 93: I thought about running this with a prom scraper deployment, but it didn't matter for this test. I figured I would leave the refactor in for future use though.

charts/agent/templates/_forwarder-config.tpl
```diff
@@ -14,7 +29,54 @@ receivers:
         endpoint: ${env:MY_POD_IP}:4317
       http:
         endpoint: ${env:MY_POD_IP}:4318
+
 processors:
+
+{{- if .Values.node.forwarder.redMetrics.enabled }}
+  # This handles schema normalization as well as moving status to attributes so it can be a dimension in spanmetrics
```
obs-gh-mattcotter, line 36: None of these comments make it past `helm template`, but I find them useful for reading the helm chart later.

integration/scripts/benchmark/log_generator.py
```diff
@@ -48,7 +48,7 @@ def _make_loggenerator_daemonset(
     rate = math.ceil(cluster_rate / float(num_nodes))
 
     container = client.V1Container(
-        image="460044344528.dkr.ecr.us-west-2.amazonaws.com/observemattc/loggenerator:latest",
+        image="observemattc/loggenerator:latest",
```
obs-gh-alexlew, line 51: is this change of image on purpose? we generally do want to host on ecr vs. from dockerhub or elsewhere

charts/agent/templates/_forwarder-config.tpl
```diff
@@ -1,5 +1,23 @@
 {{- define "observe.daemonset.forwarder.config" -}}
 
+{{- $spanmetricsResourceAttributes := (list "service.namespace" "service.version" "deployment.environment" "k8s.pod.name" "k8s.cluster.uid" "k8s.namespace.name") -}}
+
+{{- if .Values.node.forwarder.redMetrics.enabled }}
```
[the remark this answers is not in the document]
obs-gh-mattcotter: It is, but I felt like `node.forwarder.REDMetrics.enabled` was awkward since we always start with lowercase. I'm open to suggestions here. I do want RED metrics to be its own section though, in case we add more options.

obs-gh-jabernet (commented):
I think we might be able to significantly siimplify the logic by sequencing the k8sattributes processor later. Nearly all the stuff we aggregate on is produced by that processor, so it would make much more sense to aggregate over what is the unique identifier for a pod, and then just enrich on the metrics path (after the connector), not the trace path.

But that requires some more experimenting to make sure we get it right, so probably best to do it as a follow up, just wanted to call it out so we don't forget

charts/agent/values.yaml
```diff
@@ -128,6 +128,9 @@ application:
     metricDropRegex: ""
     # metrics to keep
     metricKeepRegex: (.*)
+  REDMetrics:
+      # -- (bool) Whether to enable generating RED metrics from spans. See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector#overview
+      enabled: false
```
obs-gh-mattcotter, line 133: I've updated this to `application.REDMetrics.enabled` after discussing with Alex.

obs-gh-alexlew (approved):

charts/agent/templates/_forwarder-config.tpl
```diff
@@ -1,5 +1,24 @@
 {{- define "observe.daemonset.forwarder.config" -}}
 
+{{- $spanmetricsResourceAttributes := (list "service.namespace" "service.version" "deployment.environment" "k8s.pod.name" "k8s.cluster.uid" "k8s.namespace.name") -}}
```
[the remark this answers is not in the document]
obs-gh-jabernet: I don't think so, it's something we added so I doubt anything in the regular Otel world is reliant on it.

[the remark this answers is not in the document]
obs-gh-mattcotter: I double checked this, and `k8s.cluster.uid` is actually used in the `k8sattributes` processor: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md
So I think no follow up is needed here.

PR #396, opened 2025-07-10
feat: add optional max span duration

obs-gh-mattcotter:
This is standard in competitor's agents, and we will need to restrict trace durations when we do tailsampling. The 1 hour default matches our current trace explorer behavior of ignoring spans longer than 1 hour.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,10 @@ bump-version:
 		fi; \
 	done
 
+.PHONY: pre-commit
+pre-commit:
+	pre-commit run --show-diff-on-failure --color=always --all-files
+
 .PHONY: validate-chart-version
 validate-chart-version:
 	@for chart in $(CHARTS); do \
--- a/charts/agent/ci/test-values.yaml
+++ b/charts/agent/ci/test-values.yaml
@@ -5,6 +5,17 @@ observe:
     create: true
     value: "fake-token"
 
+application:
+  prometheusScrape:
+    enabled: true
+    independentDeployment: true
+
+node:
+  forwarder:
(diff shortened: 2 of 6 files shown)


charts/agent/ci/test-values.yaml
```diff
@@ -5,6 +5,17 @@ observe:
...
+application:
+  prometheusScrape:
+    enabled: true
+    independentDeployment: true
+
+node:
+  forwarder:
+    enabled: true
+    metrics:
+      outputFormat: otel
```
obs-gh-mattcotter, line 17: Improving the test by setting our recommended values.

Makefile
```diff
@@ -87,6 +87,10 @@ bump-version:
 		fi; \
 	done
 
+.PHONY: pre-commit
+pre-commit:
```
obs-gh-mattcotter, line 91: copying this QoL change from a different PR that will be merged later

charts/agent/ci/test-values.yaml
obs-gh-jabernet: In addition we can btw test all kind of varitions in the agent-specific testing logic I found. I started something here: https://github.com/observeinc/helm-charts/compare/main...jabernet/non-default-testing but haven't gotten round to productionize yet

obs-gh-jabernet (approved):

charts/agent/ci/test-values.yaml
obs-gh-mattcotter: Nice! Thanks for starting this!

PR #409, opened 2025-08-26
feat: Add deployment.environment support

obs-gh-jabernet:
Populated deployment.environment.name globally for a cluster, analogous to cluster name & id.


charts/agent/values.yaml
```diff
@@ -20,7 +20,12 @@ observe:
 
 cluster:
   # name given to your cluster
+  # popuplates to k8s.cluster.name per https://opentelemetry.io/docs/specs/semconv/resource/k8s/#cluster
   name: observe-agent-monitored-cluster
+  deploymentEnvironment:
+    # deployment environment cluster runs in (e.g. testing, production, etc)
```
obs-gh-mattcotter, line 26: If you add `--` then the comment will be added to the readme; ie:
```suggestion
    # -- deployment environment cluster runs in (e.g. testing, production, etc)
```

charts/agent/templates/_config-processors.tpl
```diff
@@ -72,6 +72,11 @@ resource/observe_common:
       {{ else -}}
       value:  ${env:OBSERVE_CLUSTER_UID}
       {{ end -}}
+    {{ if .Values.cluster.deploymentEnvironment.name }}
+    - key: deployment.environment.name
+      action: upsert
```
obs-gh-mattcotter, line 77: Should we use `insert` instead in case some pods have explicit overrides?

charts/agent/values.yaml
obs-gh-jabernet: Neat!

charts/agent/templates/_config-processors.tpl
obs-gh-jabernet: I was considering that but I would prefer we don't allow this currently and uniformely set the environment unless some customer really asks for this. It seems more likely to cause issues than really be a use case that there's a cluster running both testing and staging for instance.

PR #410, opened 2025-08-29
feat: update to latest observe-agent version

obs-gh-mattcotter:


charts/agent/templates/_cluster-events-config.tpl
```diff
@@ -451,6 +451,4 @@ service:
         processors: [memory_limiter, batch, resource/observe_common, filter/cluster, transform/cluster]
         exporters: [{{ join ", " $logsClusterExporters }}]
 
-{{- include "config.service.telemetry" . | nindent 2 }}
```
obs-gh-mattcotter, line 454: This is now fully handled in the observe-agent configmap

charts/agent/templates/_node-logs-metrics-config.tpl
```diff
@@ -79,13 +79,6 @@ receivers:
       - pod
       - container
     metrics:
-      # Disable deprecated metrics
```
obs-gh-mattcotter, line 82: These are removed from the receiver; they are replaced by the `metricstransform` below

charts/agent/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.68.0
+version: 0.69.0
 appVersion: "2.5.0"
```
obs-gh-alexlew, line 6: lets bump this too to the observe-agent version

obs-gh-mattcotter: Done!

PR #420, opened 2025-09-10
fix: Treat "nil" ready replicas as 0

obs-gh-enricogiorio:
We currently set the status of various workload objects to "Unhealthy" when readyReplicas != desiredReplicas. However, sometimes readyReplicas could be omitted (instead of being set to 0) if no replicas are ready.

At the same time, in particular cases, one could set desiredReplicas to 0 (for whatever reason).

In this case, with the current logic, the object would always be considered "Unhealthy" since nil != 0.

Treat nil as 0 to consider objects defined with desiredReplicas == 0 always "Healthy".

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.70.2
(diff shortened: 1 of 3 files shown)


obs-gh-mattcotter (approved):
lgtm but I will defer to @obs-gh-alexlew

PR #463, opened 2026-03-16
chore: ignore examples directory in orca scan

obs-gh-mattcotter:
This covers our top level examples dir and the upcoming chart examples (see: #462)

### DIFF
--- a/.github/workflows/orca.yaml
+++ b/.github/workflows/orca.yaml
@@ -30,6 +30,7 @@ jobs:
           path:
             # scanning the entire repository
             "."
+          exclude_paths: "examples,charts/*/examples"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/orca.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
           path:
             # scanning the entire repository
             "."
+          exclude_paths: "charts/*/examples"
```
obs-gh-abhishekrao: What about the root `examples` directory?

obs-gh-mattcotter: Good point, added!

obs-gh-abhishekrao (approved):

PR #465, opened 2026-04-03
feat: update observe-agent, kube-cluster-info, and kube-state-events image versions

obs-gh-mattcotter:


obs-gh-mattcotter:
abandoning until we further validate the updated kube-state-events changes

PR #466, opened 2026-04-08
chore: add examples to exclude_paths for Orca FS Scan

obs-gh-mattcotter:


obs-gh-mattcotter:
I missed this one job and am hoping that after this the examples will really be excluded. Running the job for my examples PR on a branch with this change succeeds.

obs-gh-abhishekrao (approved):

PR #483, opened 2026-04-29
feat: shard prometheus-scraper via OpenTelemetry Target Allocator

sfc-gh-zeningchen:
## Summary

Adds an opt-in Prometheus Target Allocator (TA) sharding path for the `prometheus-scraper` Deployment. Once enabled, a standalone TA discovers targets and distributes them across scraper replicas via consistent-hashing, so the Deployment can scale horizontally without duplicating metrics.

Opt in with:

```yaml
application:
  prometheusScrape:
    independentDeployment: true
    targetAllocator:
      enabled: true
prometheus-scraper:
  replicaCount: 2   # or more
```

## Topology: before (no Target Allocator)

With a single replica, both receivers watch the K8s API directly and scrape all matching targets independently:

```
                    ┌── Kubernetes API ──────┐
                    │   pods, nodes, ...     │
                    └─────┬──────────┬───────┘
                        watch       watch
                           │           │
                           ▼           ▼
    ┌──────────────────────────────────────────────────┐
    │      prometheus-scraper   (replicaCount = 1)     │
    │                                                  │
    │  prometheus/pod_metrics                          │
    │    inline scrape_configs ──────▶ scrapes ALL pods│
    │                                                  │
    │  prometheus/cadvisor                             │
    │    inline scrape_configs ──────▶ scrapes ALL nodes│
    └──────────────────────────┬───────────────────────┘
                               ▼
                        Observe (PRW exporter)
```

Scaling `replicaCount` to N causes every replica to independently scrape all targets — N× duplication:

```
                    ┌── Kubernetes API ──────┐
                    │   pods, nodes, ...     │
                    └──┬───────┬───────┬─────┘
                     watch   watch   watch
                       │       │       │
            ┌──────────┘       │       └──────────┐
            ▼                  ▼                  ▼
    ┌──────────────┐   ┌──────────────┐   ┌──────────────┐
    │  replica 1   │   │  replica 2   │   │  replica N   │
    │  pod_metrics │   │  pod_metrics │   │  pod_metrics │
    │  cadvisor    │   │  cadvisor    │   │  cadvisor    │
    └──────┬───────┘   └──────┬───────┘   └──────┬───────┘
           │    scrapes ALL   │    scrapes ALL    │
           └──────────────────┴──────────┬────────┘
                                         ▼
                                  Observe (PRW exporter)
                                  receives N× every metric  ← problem
```

## Topology: after (with Target Allocator)

```
                    ┌── Kubernetes API ──────┐
                    │   pods, nodes, ...     │
                    └─────┬──────────┬───────┘
                       watch       watch
                          │           │
              ┌───────────┘           └───────────┐
              ▼                                   ▼
    ┌─────────────────────┐             ┌─────────────────────┐
    │ TA (pod-metrics)    │             │ TA (cadvisor)       │
    │ owns: pod-metrics   │             │ owns: kubernetes-   │
    │ allocation:         │             │   nodes-cadvisor    │
    │  consistent-hashing │             │ allocation:         │
    │                     │             │  consistent-hashing │
    └──────────┬──────────┘             └──────────┬──────────┘
               │                                   │
               │    GET /jobs/.../targets?collector_id=<pod-name>
               ▼                                   ▼
    ┌────────────────────────────────────────────────────────┐
    │        prometheus-scraper   (replicaCount = N)         │
    │                                                        │
    │  ┌─ replica 1 ──────────┐    ┌─ replica 2 ──────────┐  │
    │  │ prom/pod_metrics ───▶ TA-pod-metrics             │  │
    │  │ prom/cadvisor    ───▶ TA-cadvisor                │  │
    │  └──────────────────────┘    └──────────────────────┘  │
    │        scrapes its              scrapes its            │
    │        assigned share           assigned share         │
    └────────────────────┬───────────────────────────────────┘
                         ▼
                  Observe (PRW exporter)
```

## Why one TA per job

The OTel prometheus receiver with `target_allocator` has no per-receiver job filter — each receiver instance polls `/scrape_configs`, reads **every** job the TA returns, and scrapes all of them. The chart has two prometheus receivers (`prometheus/pod_metrics` and `prometheus/cadvisor`) because their pipelines differ in processor chain (the pod-metrics pipeline drops `service.name`; each tags metrics with its own source attribute). A single shared TA would make each receiver scrape both jobs:

```
  ┌─ one shared TA ───────────┐
  │ jobs: pod-metrics         │   /scrape_configs returns BOTH jobs
  │       cadvisor            │ ────────────────────────────────────┐
  └───────────────────────────┘                                     │
                                                                    ▼
                          ┌─────────────────────────────────────────────────┐
                          │ scraper pod                                     │
                          │   receiver pod_metrics  scrapes pod + cadvisor  │ ← 2×
                          │   receiver cadvisor     scrapes pod + cadvisor  │ ← 2×
                          └─────────────────────────────────────────────────┘
```

Every metric flows through both pipelines → 2× duplication per pod, before replica-level sharding even applies.

Isolating each job to its own TA keeps the `/scrape_configs` response to exactly one job per receiver, which is all the receiver sees, which is all it scrapes.

## What's in this PR

- `charts/agent/templates/prometheus-target-allocator.yaml` — ranges over a list of per-job TA variants (`pod-metrics` always; `cadvisor` when `node.metrics.cadvisor.enabled`), emitting a full TA stack per variant (ServiceAccount, ClusterRole, ClusterRoleBinding, ConfigMap, Service, Deployment).
- `charts/agent/templates/_config-receivers.tpl` — both `pod_metrics` and `cadvisor` receivers now branch on TA: TA mode emits a `target_allocator` block pointing at the matching TA Service; non-TA mode keeps the existing inline `scrape_configs` path.
- `charts/agent/templates/_helpers.tpl` — TA service-name helper now takes a `suffix` so each variant gets a unique DNS name.
- `charts/agent/templates/prometheus-scraper-configmap.yaml` — fails at `helm template` time if `targetAllocator.enabled` is set without `independentDeployment`.
- `charts/agent/values.yaml` — new `application.prometheusScrape.targetAllocator` block (`enabled`, `interval`, `image`, `resources`).
- `charts/agent/ci/prometheus-ta-values.yaml`, `test/verify_agent_prometheus_ta_template.sh`, `Makefile`, `.github/workflows/test.yml` — render-time checks that both TAs deploy, each TA's ConfigMap contains only its own scrape job, each receiver points at its own TA, and the non-TA path still renders cleanly.
- `charts/agent/README.md` — docs for the new flag.

## Status

**Draft / WIP** while we decide between shipping this as-is vs. a single-receiver refactor (see Alternatives). Works end-to-end locally.

## Test plan

- [x] `helm lint` clean on both the new TA CI values and the existing defaults
- [x] `helm template` verify script passes (both TAs rendered, correct endpoint wiring, no cross-contamination in TA ConfigMaps, non-TA path unchanged)
- [x] Local minikube install end-to-end with the real TA image + cadvisor sharding + probe pods (results below)
- [ ] Multi-node real-cluster soak: cadvisor target distribution across >1 node, scraper churn behavior over a few hours
- [ ] Decide design direction (see Alternatives)

## Local smoke test (minikube)

Sensitive identifiers redacted.

<details>
<summary>Setup commands</summary>

```sh
# Stop any host-level observe-agent
sudo systemctl stop observe-agent

# Fresh minikube
minikube start --driver=docker --cpus=4 --memory=6g

# Namespace + credentials
kubectl create namespace observe
kubectl -n observe create secret generic agent-credentials \
  --from-literal=OBSERVE_TOKEN='<REDACTED>'
kubectl annotate secret agent-credentials -n observe \
  meta.helm.sh/release-name=observe-agent \
  meta.helm.sh/release-namespace=observe
kubectl label secret agent-credentials -n observe \
  app.kubernetes.io/managed-by=Helm

# Install from local chart with TA + cadvisor + 2 scraper replicas
helm install observe-agent ./charts/agent -n observe \
  --set observe.collectionEndpoint.value="https://<customer-id>.collect.observe-eng.com/" \
  --set cluster.name="observe-agent-monitored-cluster" \
  --set node.containers.logs.enabled=true \
  --set node.forwarder.enabled=true \
  --set node.forwarder.metrics.outputFormat=otel \
  --set application.REDMetrics.enabled=true \
  --set agent.config.global.fleet.enabled=true \
  --set application.prometheusScrape.enabled=true \
  --set application.prometheusScrape.independentDeployment=true \
  --set application.prometheusScrape.targetAllocator.enabled=true \
  --set node.metrics.cadvisor.enabled=true \
  --set prometheus-scraper.replicaCount=2
```

</details>

### All pods Ready

```
NAME                                                      READY   STATUS
observe-agent-cluster-events-<hash>                       1/1     Running
observe-agent-cluster-metrics-<hash>                      1/1     Running
observe-agent-forwarder-agent-<hash>                      1/1     Running
observe-agent-monitor-<hash>                              1/1     Running
observe-agent-node-logs-metrics-agent-<hash>              1/1     Running
observe-agent-prometheus-scraper-<hash>-<a>               1/1     Running   # replica 1
observe-agent-prometheus-scraper-<hash>-<b>               1/1     Running   # replica 2
observe-agent-prometheus-ta-cadvisor-<hash>               1/1     Running
observe-agent-prometheus-ta-pod-metrics-<hash>            1/1     Running
```

### 1. Per-TA job isolation

```sh
kubectl -n observe port-forward svc/observe-agent-prometheus-ta-pod-metrics 18080:8080 &
kubectl -n observe port-forward svc/observe-agent-prometheus-ta-cadvisor   18081:8080 &
curl -s http://localhost:18080/scrape_configs | jq 'keys'
curl -s http://localhost:18081/scrape_configs | jq 'keys'
```

Result:

```
pod-metrics TA  /scrape_configs  →  ["pod-metrics"]
cadvisor TA     /scrape_configs  →  ["kubernetes-nodes-cadvisor"]
```

No cross-contamination — if either TA advertised both jobs, each receiver would scrape both.

### 2. Cadvisor sharding (1 node, 2 scrapers)

```sh
for id in $(kubectl get pods -n observe -l app.kubernetes.io/name=prometheus-scraper \
              -o jsonpath='{.items[*].metadata.name}'); do
  echo "$id:"
  curl -s "http://localhost:18081/jobs/kubernetes-nodes-cadvisor/targets?collector_id=$id" \
    | jq '.[].targets'
done
```

Result:

```
scraper-<a>  →  [["192.168.49.2:10250"]]
scraper-<b>  →  []
```

The single node is owned by exactly one scraper — the other gets an empty list. Cadvisor is scraped exactly once regardless of `replicaCount`.

### 3. Pod-metrics sharding with synthetic targets

Deploy 8 probe pods (each has a container port named `metrics` so the default `portKeepRegex: .*metrics` keeps them):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata: { name: probe-targets, namespace: probe }
spec:
  replicas: 8
  selector: { matchLabels: { app: probe-target } }
  template:
    metadata: { labels: { app: probe-target } }
    spec:
      containers:
        - name: probe
          image: registry.k8s.io/pause:3.9
          ports: [{ name: metrics, containerPort: 8080 }]
```

Query the pod-metrics TA:

| Step | Scrapers | Probes | Allocation | Notes |
|---|---|---|---|---|
| Initial | 2 | 8 | 4 / 4 | Perfectly even split |
| Scale scrapers to 3 | 3 | 8 | 4 / 4 / **0** | New replica sits idle |
| Scale probes to 12 | 3 | 12 | 6 / 5 / 1 | New replica picks up its share on next allocation |

Raw target count from the TA was 77 (every pod in the cluster) → `targets_remaining` was 8, then 12 after relabel filtering. The chart's rules correctly drop `kube-system` + the observe-agent's own pods (they opt out via `observeinc_com_scrape: false`).

### 4. End-to-end data flow — scrapers actually emit their assigned targets

The verify script proves that manifests render correctly; this test proves that the wired-up system *actually scrapes* and emits metrics from the assigned targets.

**Setup — install with the debug exporter wired into every pipeline:**

```sh
helm install observe-agent ./charts/agent -n observe \
  ...                                                       # (same as earlier, plus:)
  --set application.prometheusScrape.targetAllocator.enabled=true \
  --set node.metrics.cadvisor.enabled=true \
  --set prometheus-scraper.replicaCount=2 \
  --set agent.config.global.debug.enabled=true \
  --set agent.config.global.debug.verbosity=detailed \
  --set agent.config.global.service.telemetry.loggingLevel=INFO
```

**Probe targets — 8 pods serving real `/metrics`:**

```sh
kubectl apply -f - <<'YAML'
apiVersion: v1
kind: Namespace
metadata: { name: probe }
---
apiVersion: apps/v1
kind: Deployment
metadata: { name: probe-targets, namespace: probe }
spec:
  replicas: 8
  selector: { matchLabels: { app: probe-target } }
  template:
    metadata: { labels: { app: probe-target } }
    spec:
      containers:
        - name: app
          image: quay.io/brancz/prometheus-example-app:v0.5.0
          ports: [{ name: metrics, containerPort: 8080 }]
YAML
```

**Port-forward both TAs for inspection:**

```sh
kubectl -n observe port-forward svc/observe-agent-prometheus-ta-pod-metrics 18080:8080 &
kubectl -n observe port-forward svc/observe-agent-prometheus-ta-cadvisor   18081:8080 &
```

#### Pod-metrics

**Command — TA allocator gauges:**

```sh
curl -s http://localhost:18080/metrics | grep '^opentelemetry_allocator_'
```

**Result:**

```
opentelemetry_allocator_collectors_discovered 2
opentelemetry_allocator_collectors_allocatable{strategy="consistent-hashing"} 2
opentelemetry_allocator_targets{job_name="pod-metrics"} 69
opentelemetry_allocator_targets_remaining 8
opentelemetry_allocator_targets_unassigned 0
opentelemetry_allocator_targets_per_collector{collector_name="...hbnj4",strategy="consistent-hashing"} 6
opentelemetry_allocator_targets_per_collector{collector_name="...v5grp",strategy="consistent-hashing"} 2
```

69 raw targets from K8s SD → 8 kept after relabel filtering → split 6/2 between the two scrapers. Imbalance is expected for small target counts with consistent-hashing; evens out as target count grows.

**Command — which probe pods each scraper actually emitted metrics for (last 70s window, one scrape cycle after allocation settled):**

```sh
for p in $(kubectl -n observe get pods -l app.kubernetes.io/name=prometheus-scraper \
             -o jsonpath='{.items[*].metadata.name}'); do
  echo "--- $p ---"
  kubectl -n observe logs "$p" --since=70s \
    | grep -oE 'service\.instance\.id: Str\(10\.244\.[0-9]+\.[0-9]+:8080\)' \
    | sort -u
done
```

**Result:**

```
--- scraper hbnj4 ---
service.instance.id: Str(10.244.0.12:8080)
service.instance.id: Str(10.244.0.13:8080)
service.instance.id: Str(10.244.0.14:8080)
service.instance.id: Str(10.244.0.15:8080)
service.instance.id: Str(10.244.0.18:8080)
service.instance.id: Str(10.244.0.19:8080)

--- scraper v5grp ---
service.instance.id: Str(10.244.0.16:8080)
service.instance.id: Str(10.244.0.17:8080)
```

**Command — overlap + union:**

```sh
A=$(kubectl -n observe logs scraper-hbnj4 --since=70s | grep -oE '10\.244\.[0-9]+\.[0-9]+:8080' | sort -u)
B=$(kubectl -n observe logs scraper-v5grp --since=70s | grep -oE '10\.244\.[0-9]+\.[0-9]+:8080' | sort -u)
comm -12 <(echo "$A") <(echo "$B")             # overlap
{ echo "$A"; echo "$B"; } | sort -u | wc -l    # union
```

**Result:**

```
overlap: (empty — no probe pod appears in both scrapers' emissions)
union:   8                  (every probe target is scraped exactly once)
```

The OTel prometheus receiver stamps `service.instance.id` as a resource attribute on every emitted metric, set to the `<ip>:<port>` that was actually scraped. So zero overlap between the two scrapers' `service.instance.id` sets proves the deduplication guarantee the whole feature is designed for: **every probe pod's `/metrics` endpoint was hit by exactly one scraper in this scrape cycle — no pod was scraped twice, and no pod was missed.**

**Command — which metric names each scraper actually emitted for the probe pods:**

```sh
for p in $(kubectl -n observe get pods -l app.kubernetes.io/name=prometheus-scraper \
             -o jsonpath='{.items[*].metadata.name}'); do
  echo "--- $p ---"
  kubectl -n observe logs "$p" --since=70s \
    | awk '/service\.name: Str\(probe-targets\)/{p=1} /^2026/{p=0} p && /^ *-> Name:/' \
    | sort -u
done
```

**Result:**

```
--- scraper hbnj4 ---
     -> Name: scrape_duration_seconds          # Prometheus meta
     -> Name: scrape_samples_post_metric_relabeling
     -> Name: scrape_samples_scraped
     -> Name: scrape_series_added
     -> Name: up
     -> Name: version                          # the probe app's own metric

--- scraper v5grp ---
     -> Name: scrape_duration_seconds
     -> Name: scrape_samples_post_metric_relabeling
     -> Name: scrape_samples_scraped
     -> Name: scrape_series_added
     -> Name: up
     -> Name: version                          # the probe app's own metric
```

`prometheus-example-app` exposes a single data metric (`version{version="v0.5.0"} 1`); it arrives in the scraper's debug exporter output with `debug_source: Str(pod_metrics)` alongside Prometheus' scrape meta-metrics. Both scrapers emit it for their own assigned subset of probe pods.

#### Cadvisor

**Caveat:** minikube is a single-node cluster, so cadvisor service discovery finds exactly **1 target**. With 1 target and 2 scrapers, consistent-hashing deterministically assigns the target to one scraper; the other correctly gets an empty list. This is the degenerate sharding case (targets < collectors), not a full sharding proof for cadvisor — a multi-node cluster would show the same split-across-scrapers behavior demonstrated above for pod-metrics. The multi-node soak is in the Test plan as a follow-up.

**Command — TA allocator gauges:**

```sh
curl -s http://localhost:18081/metrics | grep '^opentelemetry_allocator_'
```

**Result:**

```
opentelemetry_allocator_targets{job_name="kubernetes-nodes-cadvisor"} 1
opentelemetry_allocator_targets_per_collector{collector_name="...hbnj4"} 1
opentelemetry_allocator_targets_per_collector{collector_name="...v5grp"} 0
```

**Command — count cadvisor metric emissions per scraper:**

```sh
for p in $(kubectl -n observe get pods -l app.kubernetes.io/name=prometheus-scraper \
             -o jsonpath='{.items[*].metadata.name}'); do
  n=$(kubectl -n observe logs "$p" --since=3m | grep -c 'container_cpu_usage_seconds_total')
  echo "$p: container_cpu_usage_seconds_total hits = $n"
done
```

**Result:**

```
scraper hbnj4: container_cpu_usage_seconds_total hits = 3   # owns the cadvisor target
scraper v5grp: container_cpu_usage_seconds_total hits = 0   # empty target list
```

Only the scraper owning the cadvisor target produces cadvisor metrics; the other produces none. The full cadvisor metric family (`container_blkio_*`, `container_cpu_*`, `container_fs_*`, `container_memory_*`, `container_network_*`, etc.) flows through the `metrics/cadvisor` pipeline as expected.

#### Summary

Across both receivers, every target is scraped exactly once; every metric hits the pipeline exactly once per scrape cycle. No duplication across scraper replicas.

### Steady-state logs

```sh
kubectl -n observe logs <scraper-pod> --since=30s | grep -iE 'error|warn'
```

No errors in steady state. Initial startup shows a few `Failed to retrieve job list: connection refused` — the scraper pod polls the TA before the TA pod is Ready. The receiver's TA manager retries on its interval and recovers; noted as a caveat below.

## Known caveats

1. **Startup race.** Scraper polls TA before TA is Ready → a few seconds of `connection refused` at boot. Self-corrects. Follow-up: initContainer that waits for the TA Service's endpoints to have a Ready pod.
2. **New scraper replicas don't receive targets until the target set changes.** Upstream TA's consistent-hashing allocator only re-runs allocation on target-change events, not collector-change events. In production this self-corrects within seconds (pods churn naturally); in a quiet test setup you have to perturb the target set to trigger reallocation. This is upstream behavior, not a chart bug — but worth documenting.
3. **TA image is distroless** — no `wget`, `curl`, or shell for in-container debugging. Use `kubectl port-forward` + host tooling.
4. **No `checksum/config` on the TA Deployment pod template.** Editing a TA ConfigMap won't trigger a rollout. Follow-up.

## Alternatives considered

**Single TA, two receivers.** The naive approach. Rejected — see "Why one TA per job" above.

**Single receiver + routing/filter processors.** Collapse the two prometheus receivers into one `prometheus/k8s_metrics`, split the resulting stream into two pipelines via `filter` processors keyed on `service.name` (which the prometheus receiver sets to the scrape job name). Matches the upstream "one receiver per collector" pattern and is where the OTel ecosystem is trending. Not in this PR because it's a larger refactor (receivers, pipelines, processors all change) and relies on an implicit contract between the prometheus receiver's label mapping and the filter rule — worth doing, but as a separate change where routing correctness can be tested in isolation.

### DIFF
--- a/charts/agent/values.yaml
+++ b/charts/agent/values.yaml
@@ -144,6 +144,22 @@ application:
     metricDropRegex: ""
     # metrics to keep
     metricKeepRegex: (.*)
+    # Target Allocator (TA-only) shards pod_metrics across multiple prometheus-scraper replicas.
+    # Requires application.prometheusScrape.independentDeployment=true. OpenTelemetry Operator is not required.
+    # Set prometheus-scraper.replicaCount >= 2 for meaningful sharding.
+    targetAllocator:
+      enabled: false
+      # How often each collector syncs scrape jobs from the Target Allocator (Prometheus receiver target_allocator.interval).
+      interval: 30s
+      image:
+        repository: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator
+        tag: "0.150.0"
+      resources:
+        requests:
+          cpu: 50m
+          memory: 128Mi
+        limits:
+          memory: 256Mi
   REDMetrics:
     # -- (bool) Whether to enable generating RED metrics from spans. See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector#overview
     enabled: false
@@ -628,6 +644,10 @@ prometheus-scraper:
       memory: 512Mi
     limits:
       memory: 512Mi
+
+  securityContext:
+    seccompProfile:
+      type: RuntimeDefault
   # ----------------------------------------- #
 
   # ----------------------------------------- #
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,9 @@ jobs:
       - name: Lint charts
         run: make lint
 
+      - name: Verify agent Prometheus Target Allocator rendering
+        run: make verify-agent-prometheus-ta
+
   test-charts:
     runs-on: ubuntu-latest
     steps:
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,10 @@ build-deps: add-repos
 		echo ; \
 	done
 
+.PHONY: verify-agent-prometheus-ta
+verify-agent-prometheus-ta:
+	test/verify_agent_prometheus_ta_template.sh
+
 .PHONY: test
 test: build-deps build-test-images
 	test/test.sh agent stack traces
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.86.1
+version: 0.87.0
 appVersion: "2.15.0"
 dependencies:
   - name: opentelemetry-collector
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.86.1](https://img.shields.io/badge/Version-0.86.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.15.0](https://img.shields.io/badge/AppVersion-2.15.0-informational?style=flat-square)
+![Version: 0.87.0](https://img.shields.io/badge/Version-0.87.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.15.0](https://img.shields.io/badge/AppVersion-2.15.0-informational?style=flat-square)
 
 Chart to install K8s collection stack based on Observe Agent
 
@@ -21,6 +21,8 @@ This service is a *single-instance deployment*. It's critical that this service 
 ## prometheus-scraper
 
 This service is a *single-instance deployment*. It's critical that this service is only a single instance since otherwise it would produce duplicate data. It is responsible for scraping pods for Prometheus metrics is configured and runs.
+
+When **`application.prometheusScrape.targetAllocator.enabled`** is `true` (requires **`application.prometheusScrape.independentDeployment`**), the chart also deploys an OpenTelemetry **Target Allocator** and configures the `prometheus/pod_metrics` receiver to shard targets across **`prometheus-scraper.replicaCount`** observe-agent pods. Set **`prometheus-scraper.replicaCount`** to at least `2` for meaningful sharding. The OpenTelemetry Operator is not required for this path. Pair the **`targetAllocator.image`** tag with your observe-agent / OTel collector version per OpenTelemetry release notes.
 
 ## forwarder
 
@@ -101,6 +103,13 @@ This service is an *OpenTelemetryCollector*, a custom resource that is managed b
 | application.prometheusScrape.namespaceDropRegex | string | `"(.*istio.*|.*ingress.*|kube-system)"` |  |
 | application.prometheusScrape.namespaceKeepRegex | string | `"(.*)"` |  |
 | application.prometheusScrape.portKeepRegex | string | `".*metrics"` |  |
+| application.prometheusScrape.targetAllocator.enabled | bool | `false` |  |
+| application.prometheusScrape.targetAllocator.image.repository | string | `"ghcr.io/open-telemetry/opentelemetry-operator/target-allocator"` |  |
+| application.prometheusScrape.targetAllocator.image.tag | string | `"0.150.0"` |  |
+| application.prometheusScrape.targetAllocator.interval | string | `"30s"` |  |
+| application.prometheusScrape.targetAllocator.resources.limits.memory | string | `"256Mi"` |  |
+| application.prometheusScrape.targetAllocator.resources.requests.cpu | string | `"50m"` |  |
+| application.prometheusScrape.targetAllocator.resources.requests.memory | string | `"128Mi"` |  |
 | cluster-events.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].key | string | `"observeinc.com/unschedulable"` |  |
 | cluster-events.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[0].operator | string | `"DoesNotExist"` |  |
 | cluster-events.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions[1].key | string | `"kubernetes.io/os"` |  |
@@ -703,6 +712,7 @@ This service is an *OpenTelemetryCollector*, a custom resource that is managed b
 | prometheus-scraper.resources.limits.memory | string | `"512Mi"` |  |
 | prometheus-scraper.resources.requests.cpu | string | `"250m"` |  |
 | prometheus-scraper.resources.requests.memory | string | `"512Mi"` |  |
+| prometheus-scraper.securityContext.seccompProfile.type | string | `"RuntimeDefault"` |  |
 | prometheus-scraper.serviceAccount.create | bool | `false` |  |
 | prometheus-scraper.serviceAccount.name | string | `"observe-agent-service-account"` |  |
 | prometheus-scraper.tolerations | list | `[]` |  |
--- a/charts/agent/README.md.gotmpl
+++ b/charts/agent/README.md.gotmpl
@@ -23,6 +23,8 @@ This service is a *single-instance deployment*. It's critical that this service 
 
 This service is a *single-instance deployment*. It's critical that this service is only a single instance since otherwise it would produce duplicate data. It is responsible for scraping pods for Prometheus metrics is configured and runs.
 
+When **`application.prometheusScrape.targetAllocator.enabled`** is `true` (requires **`application.prometheusScrape.independentDeployment`**), the chart also deploys an OpenTelemetry **Target Allocator** and configures the `prometheus/pod_metrics` receiver to shard targets across **`prometheus-scraper.replicaCount`** observe-agent pods. Set **`prometheus-scraper.replicaCount`** to at least `2` for meaningful sharding. The OpenTelemetry Operator is not required for this path. Pair the **`targetAllocator.image`** tag with your observe-agent / OTel collector version per OpenTelemetry release notes.
+
 ## forwarder
 
 This service is a *daemonset* which means it runs on every node in the cluster. It is responsible for receiving telemetry from the other services, specifically via an OTLP receiver and forwarding it to Observe. It can be used as the target for various instrumentation SDK's and clients as well. See [here](https://docs.observeinc.com/en/latest/content/observe-agent/ConfigureApplicationInstrumentation.html) for more details.
--- a/charts/agent/ci/prometheus-ta-values.yaml
+++ b/charts/agent/ci/prometheus-ta-values.yaml
@@ -1,0 +1,60 @@
+# Same as test-values.yaml but enables Target Allocator sharding for pod_metrics.
+# chart-testing runs `helm template` with each ci/*-values.yaml independently.
+observe:
+  collectionEndpoint:
+    value: "http://test-stack-collector.testing.svc.cluster.local:8080"
+  token:
+    create: true
+    value: "fake-ds:fake-token"
+
+application:
+  prometheusScrape:
+    enabled: true
+    independentDeployment: true
+    targetAllocator:
+      enabled: true
+  REDMetrics:
+    enabled: true
+
+node:
+  forwarder:
+    enabled: true
+    metrics:
+      outputFormat: otel
+      convertCumulativeToDelta: true
+  metrics:
+    cadvisor:
+      enabled: true
+
+gatewayDeployment:
+  enabled: true
+  traceSampling:
+    enabled: true
+
+agent:
+  config:
+    global:
+      fleet:
+        enabled: true
+        interval: 10s
+        configInterval: 10m
+
+cluster:
+  namespaceOverride:
+    value: "testing"
+
+cluster-events:
+  namespaceOverride: "testing"
+cluster-metrics:
+  namespaceOverride: "testing"
+forwarder:
+  namespaceOverride: "testing"
+monitor:
+  namespaceOverride: "testing"
+node-logs-metrics:
+  namespaceOverride: "testing"
+prometheus-scraper:
+  namespaceOverride: "testing"
+  replicaCount: 2
+gateway:
+  namespaceOverride: "testing"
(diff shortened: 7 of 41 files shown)


charts/agent/values.yaml
```diff
@@ -144,6 +144,22 @@ application:
...
+    # Target Allocator (TA-only) shards pod_metrics across multiple prometheus-scraper replicas.
+    # Requires application.prometheusScrape.independentDeployment=true. OpenTelemetry Operator is not required.
+    # Set prometheus-scraper.replicaCount >= 2 for meaningful sharding.
+    targetAllocator:
+      enabled: false
+      # How often each collector syncs scrape jobs from the Target Allocator (Prometheus receiver target_allocator.interval).
+      interval: 30s
+      image:
+        repository: ghcr.io/open-telemetry/opentelemetry-operator/target-allocator
+        tag: "0.118.0"
```
obs-gh-abhishekrao: This image may be a bit old. I see 0.150.0 is the latest here https://github.com/open-telemetry/opentelemetry-operator/pkgs/container/opentelemetry-operator%2Ftarget-allocator

sfc-gh-zeningchen: Bumped to 0.150.0, thanks!

sfc-gh-zeningchen:
this is just a prototype, implemented it in #485 , thus closing this pr

PR #485, opened 2026-05-04
feat: shard prometheus-scraper via opentelemetry-target-allocator

sfc-gh-zeningchen:
## Summary

Enables horizontal sharding of the prometheus-scraper via the upstream
opentelemetry-target-allocator (TA). Default off
(`application.prometheusScrape.targetAllocator.enabled=false`); non-TA path
renders identically to main.

## Changes

- **Merge prometheus receivers**: `prometheus/pod_metrics` +
  `prometheus/cadvisor` → one `prometheus/k8s_metrics` feeding one pipeline.
  `debug_source` stamped per-datapoint by `transform/set_debug_source` (keys
  off `service.name`). Eliminates the fan-out cost that dominated TA-path
  memory.
- **Deploy TA subchart** (alias `prometheus-ta`, v0.128.1).
  `prometheus/k8s_metrics` switches to `target_allocator:` with
  `collector_id = ${OTEL_K8S_POD_NAME}` so consistent-hashing distributes
  targets across replicas. The TA's `targetallocator.yaml` is rendered from
  the same scrape-job partials as the non-TA path.
- **Gates / fixes** required for the TA path to be safe:
  - `metric_relabel` drop rule emitted only when `metricDropRegex` is
    non-empty (empty round-tripped through TA as `(.*)`, silently dropping
    every user metric).
  - `targetAllocator.enabled=true && cadvisor.separate_pipeline=true` blocked
    at render time (split receivers leave half the targets unsharded).
  - `replicaCount > 1 && targetAllocator.enabled=false` blocked (every
    replica would otherwise scrape every target and emit duplicates).

## Sizing

Above ~1000 targets at typical ~1500 series/target cardinality:
`replicaCount = ceil(targets / 110)`. EKS load test to 2000 targets / 18
replicas confirms the rule; TA itself runs comfortably on chart defaults
(50m / 256 MiB).

## Validation

Side-by-side dual-ship at the same eng datastream:
- 79 distinct metrics on both paths, 0 samples missing `debug_source` on the
  new path (legacy path drops 260 `target_info` samples).
- Short-lived job capture (`interval=1s`) within binomial noise across all
  lifetime buckets from 5s to 120s.
- Non-TA default render diff vs `main` = 0 lines.


charts/agent/values.yaml
```diff
@@ -144,6 +144,11 @@ application:
     metricDropRegex: ""
     # metrics to keep
     metricKeepRegex: (.*)
+    # Merge pod-metrics and cadvisor into a single receiver with filter processors to split pipelines.
+    # Requires application.prometheusScrape.independentDeployment=true.
+    # Sharding across replicas via Target Allocator will be introduced in a follow-up PR.
+    targetAllocator:
+      enabled: false
```
obs-gh-mattcotter, line 157: This value seems a bit misleading. I think it would be better to control this through something like `node.metrics.cadvisor.separate_pipeline`. We can default it to true, but ignore it later when TA is turned on

sfc-gh-zeningchen: thanks, that's a good point! I just renamed it, and I just renamed it back to `targetAllocator.enabled` in the next PR

charts/agent/templates/_config-pipelines.tpl
```diff
@@ -41,6 +41,31 @@ metrics/spanmetrics:
 {{- end -}}
 
 {{- define "config.pipelines.prometheus_scrapers" -}}
+{{- $merged := and .Values.application.prometheusScrape.independentDeployment (not .Values.node.metrics.cadvisor.separate_pipeline) }}
```
obs-gh-mattcotter: Why is this only merged when `independentDeployment` is set?

charts/agent/values.yaml
```diff
@@ -65,6 +65,11 @@ node:
       excludeMountPoints: '["/dev/*","/proc/*","/sys/*","/run/k3s/containerd/*","/var/lib/docker/*","/var/lib/kubelet/*","/snap/*"]'
     cadvisor:
       enabled: false
+      # Merge cadvisor and pod-metrics into a single OpenTelemetry pipeline.
```
obs-gh-mattcotter: Please add the `--` at the start of this comment so it gets put into the README docs

obs-gh-mattcotter (commented):
Two small comments, otherwise lgtm!

charts/agent/templates/_config-pipelines.tpl
sfc-gh-zeningchen: thanks for calling it out! I've make it available even when independentDeployment=false so that people who wants to merge these receivers inside `cluster-metrics` can also use that.

charts/agent/values.yaml
sfc-gh-zeningchen: thanks! just added it

charts/agent/templates/_config-receivers.tpl
```diff
@@ -138,6 +146,36 @@ prometheus/cadvisor:
...
+  both jobs. Filter processors in the pipelines split the stream by service.name.
+  NOTE: uses $$1 (OTel expands $$ → $) unlike the standalone cadvisor receiver.
+*/}}
+{{- define "config.receivers.prometheus.k8s_metrics" -}}
+prometheus/k8s_metrics:
+  config:
+    scrape_configs:
+{{ include "observe.prometheus.pod_metrics.scrape_job" . | nindent 4 }}
+{{- if .Values.node.metrics.cadvisor.enabled }}
+    - job_name: 'kubernetes-nodes-cadvisor'
```
obs-gh-mattcotter: I think it's worth sharing the config for this job too just in case we need to make changes.

obs-gh-mattcotter (approved):
One more comment, but then lgtm!

charts/agent/templates/_config-receivers.tpl
sfc-gh-zeningchen: sure, updated it, thanks for the review!

PR #486, opened 2026-05-05
feat: deploy upstream opentelemetry-target-allocator subchart for TA sharding

sfc-gh-zeningchen:
## Summary

Adds the upstream `opentelemetry-target-allocator` v0.128.0 (TA binary 0.151.0) as a subchart (alias `prometheus-ta`) and switches `prometheus/k8s_metrics` to a `target_allocator:` block. Each scraper identifies itself with `collector_id = ${OTEL_K8S_POD_NAME}` so consistent-hashing distributes targets across replicas.

Also includes the fixes/gates needed for the TA path to be usable:

- **`metric_relabel` guard** — empty `metricDropRegex` round-tripped through TA as `(.*)`, silently dropping every user-emitted metric (synthetic `up`/`scrape_*` survived, masking it). Drop rule now only renders when the regex is non-empty.
- **TA poll interval** surfaced as a values field (default `30s`, OTel prometheus receiver's documented default). Worst-case capture-rate gap is `T_TA / (2T)` — 25 pp at defaults, 0.83 pp at `1s`.
- **Merged-pipeline gate** — `targetAllocator.enabled=true && cadvisor.separate_pipeline=true` is blocked at render time. TA's consistent-hashing operates per-receiver; split receivers leave half the targets unsharded.
- **`prometheus-scraper.replicaCount` gate** — `replicaCount>1 && targetAllocator.enabled=false` is blocked. Without TA every replica scrapes the same targets and produces N× duplicate samples.

The TA's `targetallocator.yaml` ConfigMap is rendered from the same scrape-job partials the non-TA path uses (single source of truth). Non-TA default render is unchanged (0-line diff).

### minikube

32 GB / 8 CPU minikube. Two releases of the chart at the same eng datastream — `observe-old` (`targetAllocator.enabled=false`) and `observe-new` (`targetAllocator.enabled=true`, `interval=1s`). Scraper memory cap = 512 MiB.

**Metric parity** — 50 × `prometheus-example-app` + traffic-gen, 5-min steady-state:

| | OLD | NEW |
|---|---:|---:|
| distinct metrics | 79 | 79 |
| total samples | 44,205 | 45,416 (+2.7%) |
| samples without `debug_source` | 260 (all `target_info`) | 0 |

All 5 application metrics arrive on both sides. Without the `metric_relabel` guard the new path would deliver 0 application samples.

**Horizontal scale-out** — 80 avalanche pods at varying `SERIES_COUNT`:

| series/pod | total | OLD(1) restarts | NEW(2) restarts | OLD mem | NEW mem (each) |
|---:|---:|---:|---:|---:|---:|
| 500 | 40 K | 0 | 0 | 394 Mi | ~390 Mi |
| 600 | 48 K | 0 | 0 | 445 Mi | ~395 Mi |
| **700** | 56 K | **3, OOMKilled** | **0** | over 512 Mi cap | ~410 Mi |

Per-series cost is ~2× legacy (one collector baseline paid per replica), so practical N=2 capacity is ~1.4–1.5×, not 2×.

**Short-lived job capture** — 8 lifetime buckets × 100 reps, `interval=1s`:

| L (s) | 5 | 10 | 15 | 30 | 45 | 60 | 90 | 120 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| OLD | 6 | 9 | 21 | 53 | 61 | 83 | 99 | 100 |
| NEW | 2 | 18 | 22 | 48 | 66 | 97 | 100 | 100 |

Within binomial noise at every bucket. Worst-case gap matches the closed-form `T_TA / (2T)` = 0.83 pp.

### EKS load test

10× → 25× `m5.4xlarge`, TA at chart defaults (50m / 256 Mi), avalanche pods at ~1550 series each, 5-min warmup + 15-min sample window.

| | 380 / 6 replicas | 2000 / 14 replicas | 2000 / 18 replicas |
|---|---:|---:|---:|
| Nodes | 10 | 25 | 25 |
| Targets/replica | 63 | 143 | 111 |
| Active series | ~589 K | ~3.1 M | ~3.1 M |
| Worst p95 mem | 407 MiB | **511 MiB** | 408 MiB |
| Headroom | 20% | **0.1%** | 20% |
| OOMKilled | 0/6 | **3/14** | 0/18 |
| Per-replica mem CV | 0.011 | 0.073 | 0.010 |
| TA p95 mem | 52 MiB | 134 MiB | 134 MiB |
| TA p95 CPU | 2.1 mcores | 9 mcores | 9 mcores |

Sizing rule: **`replicaCount = ceil(targets / 110)`** for clusters above ~1000 targets at typical (~1500 series/target) cardinality. The 14-replica row violates the rule (143 targets/replica) and shows what failure looks like — 3/14 OOMKilled, worst survivor pinned at the cap. TA itself is comfortable on chart defaults across all runs.


charts/agent/templates/_config-receivers.tpl
```diff
@@ -147,33 +178,17 @@ prometheus/cadvisor:
...
-        - role: node
-      relabel_configs:
-        - target_label: __address__
-          replacement: kubernetes.default.svc:443
-        - source_labels: [__meta_kubernetes_node_name]
-          regex: (.+)
-          target_label: __metrics_path__
-          replacement: /api/v1/nodes/$$1/proxy/metrics/cadvisor
-{{- end }}
+  target_allocator:
```
obs-gh-mattcotter, line 169: I think we should offer three modes for scraping both pod and CAdvisor metrics:

1. no TA, two separate pipelines (same as today)
2. no TA, combined pipeline (enabled by your other PR)
3. yes TA, combined pipeline

This PR is adding mode 3, but seems to be getting rid of 2. Can we keep that option as well?

charts/agent/values.yaml
```diff
@@ -615,9 +630,39 @@ cluster-metrics:
...
+prometheus-ta:
+  # !!! IMPORTANT !!! This needs to have same value as namespaceOverride in cluster above
+  namespaceOverride: "observe"
+  configMap:
+    create: false
+    existingName: "prometheus-ta"
+  targetAllocator:
+    resources:
+      requests:
+        cpu: 50m
```
obs-gh-mattcotter, line 634: This is pretty low; let's make sure to benchmark this and tune these values.

charts/agent/values.yaml
```diff
@@ -615,9 +630,39 @@ cluster-metrics:
...
+        memory: 128Mi
```
obs-gh-mattcotter: Best practice is to set memory request = memory limit unless there's a good reason not to. Here's a good post about this: https://home.robusta.dev/blog/kubernetes-memory-limit
But the summary is that having a higher memory limit than request means you can over-allocate your node.

sfc-gh-zeningchen: thanks, good to learn that! I just updated the request to match the memory limit

charts/agent/templates/_config-receivers.tpl
sfc-gh-zeningchen: thanks for catching that! I've added a branch on the template to keep the old scraper setting for mode 2

charts/agent/values.yaml
sfc-gh-zeningchen: that value was randomly decided but it seems like TA is a low-CPU job, during the load test (380 pod-metrics targets across 6 sharded scrapers on EKS) TA p95 CPU was 2 mcores(max 2.79). But on the other hand, the job is spiky(scraper asks for jobs every 30 seconds), and the CPU usage was reported as an average number, not sure if that would change the meaning of what we saw here...

obs-gh-mattcotter: Oh interesting. I guess it just isn't doing that much work? Something like: 1. issue k8s API request and filter 2. make a mapping between targets and scrapers 3. listen to inbound requests from the scrapers (1req/scraper/min) and serve the prom URLs 4. update mapping if a scraper joins or leaves

I guess it's possible that the CPU usage is truly quite low. You could try to shorten the poll interval of the k8s metrics API and see if you can catch the pod at a busy time? Otherwise I'm happy testing this out live and updating if we need to.

sfc-gh-zeningchen: I did some research on how target allocator did service discovery and realized that it maintains a long-lived HTTP connection with k8s apiserver, thus the update to pods/nodes are pushed to ta at real-time. But I did update the scraper's polling interval from ta from `30s` to `1s` and ran with the same setting to make sure frequent polling doesn't overwhelm the ta, but it turned out that even with `1s` interval the peak CPU usage was around `5m`, here's the detailed comparison:
```
  ┌───────────────────────┬──────────────┬─────────────┬──────────────────┐
  │                       │ 30s interval │ 1s interval │        Δ         │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ TA p100 CPU           │  2.79 mcores │ 4.85 mcores │ +74%             │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ TA p95 CPU            │         2.11 │        3.63 │ +72%             │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ TA median CPU         │         1.07 │        2.07 │ +93%             │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ TA p95 memory         │       52 MiB │      46 MiB │ flat (run noise) │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ Worst scraper p95 mem │      407 MiB │     404 MiB │ flat (sanity)    │
  ├───────────────────────┼──────────────┼─────────────┼──────────────────┤
  │ Worst scraper p95 CPU │       128 mc │      109 mc │ flat (sanity)    │
  └───────────────────────┴──────────────┴─────────────┴──────────────────┘
```
Maybe we can test it out live to confirm.

charts/agent/templates/_config-receivers.tpl
```diff
@@ -104,28 +104,28 @@
...
-    scrape_configs:
-{{ include "observe.prometheus.pod_metrics.scrape_job" . | nindent 4 }}
-{{- end -}}
-
 {{/*
   Prometheus scrape_configs entry for cadvisor (Kubernetes node SD via the API-server proxy).
-  Shared across the standalone prometheus/cadvisor receiver and the merged
-  prometheus/k8s_metrics receiver. Defines the backreference as $1; OTel collector
-  receivers must apply `replace "$1" "$$1"` to escape it through OTel's config-expansion.
+  Used by the parent-rendered TA ConfigMap. Uses $1 (not $$1) — the TA reads this YAML
```
obs-gh-mattcotter: This comment is a bit confusing; what is `$1` note saying?

charts/agent/templates/_config-receivers.tpl
```diff
@@ -155,19 +162,26 @@ prometheus/cadvisor:
 {{ end }}
 
 {{/*
-  Merged receiver for the TA-enabled path. Combines pod-metrics and cadvisor jobs into
-  a single receiver so that one Target Allocator (introduced in a follow-up PR) can serve
-  both jobs. Filter processors in the pipelines split the stream by service.name.
-  NOTE: uses $$1 (OTel expands $$ → $) unlike the standalone cadvisor receiver.
+  Merged receiver for the TA-enabled path. Polls the upstream Target Allocator
```
obs-gh-mattcotter: I think this comment is wrong. It's the receiver for the merged path either with or without the TA. It also seems pretty verbose. The new comments throughout this diff could do with some copy edits for clarity.

obs-gh-mattcotter (commented):
Looks good! A few requests for comment cleanup, otherwise let's ship it!

charts/agent/templates/_config-receivers.tpl
sfc-gh-zeningchen: thanks! I've also updated the other comments trying to make it simpler

charts/agent/templates/_config-receivers.tpl
sfc-gh-zeningchen: it is regexp extracted node name, updated the comment to simplify and clarify it

charts/agent/Chart.lock
```diff
@@ -8,6 +8,9 @@ dependencies:
 - name: opentelemetry-collector
   repository: https://open-telemetry.github.io/opentelemetry-helm-charts
   version: 0.143.0
+- name: opentelemetry-target-allocator
+  repository: https://open-telemetry.github.io/opentelemetry-helm-charts
+  version: 0.128.1
```
obs-gh-mattcotter, line 13: The other chart versions have since been updated, please make sure this is in sync with them. Also a comment about how to synchronize this version with the `opentelemetry-collector` chart version would be useful for the future!

obs-gh-mattcotter (approved):
Looks great, thanks! One more small comment, but after addressing feel free to merge

charts/agent/Chart.lock
sfc-gh-zeningchen: rebased and added comments to Chart.yaml, thanks for the review!

PR #488, opened 2026-05-08
fix: removes emission of target_info Prometheus metric as Observe does not use it

obs-gh-francisco:


obs-gh-scottsolmonson:
Confirmed as working...

obs-gh-scottsolmonson (approved):

PR #489, opened 2026-05-11
feat: add emitPrometheusTargetInfoMetrics config value, default to false so target_info metrics will no longer be written

obs-gh-mattcotter:
Observe doesn't use the value, so we can ignore by default.

Replaces: #488

### DIFF
--- a/.orca-security.yaml
+++ b/.orca-security.yaml
@@ -4,3 +4,10 @@ global:
     - "./examples/**"
     - "./integration/scripts/benchmark/**"
     - "./charts/stack/templates/tests/**"
+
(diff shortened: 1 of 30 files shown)


obs-gh-francisco (approved):

obs-gh-abhishekrao (approved):
LGTM. I'm guessing we also need to port this to the observe-agent repo.

obs-gh-mattcotter:
> LGTM. I'm guessing we also need to port this to the observe-agent repo.

Yes, I'm making that update here: https://github.com/observeinc/observe-agent/pull/341

PR #505, opened 2026-05-28
feat: support ServiceMonitor/PodMonitor CRD discovery via Target Allocator

sfc-gh-zeningchen:
## Summary

Adds ServiceMonitor / PodMonitor CRD discovery on the existing Target
Allocator path. With Prometheus Operator CRDs in the cluster, TA discovers
scrape targets natively — useful for clusters whose SMs use non-`*metrics`
port names that the chart's static keep-relabel rejects.

## New values

Under `application.prometheusScrape.targetAllocator`:

| Value | Default | Purpose |
|---|---|---|
| `prometheusCR.enabled` | `false` | Toggle CRD discovery |
| `prometheusCR.serviceMonitorSelector` | `{}` | SM label selector (empty = all) |
| `prometheusCR.podMonitorSelector` | `{}` | PM label selector (empty = all) |
| `prometheusCR.serviceMonitorNamespaceSelector` | `{}` | SM namespace selector (empty = all) |
| `prometheusCR.podMonitorNamespaceSelector` | `{}` | PM namespace selector (empty = all) |
| `includeStaticScrapeConfigs` | `true` | Keep chart's static pod-metrics + cadvisor jobs alongside CRD targets |

Render-time guard rejects `prometheusCR.enabled=false && includeStaticScrapeConfigs=false`.

## Backward compatibility

Default render unchanged — both new flags default such that existing TA
users keep current behavior. Diff vs main on default values = 0 lines.

## Test

Minikube + Prometheus Operator CRDs + `prometheus-example-app` exposed on
container port name `https` (deliberately not matching `*metrics`, so the
chart's static keep-relabel would reject it). Installed with:

```yaml
application:
  prometheusScrape:
    enabled: true
    independentDeployment: true
    targetAllocator:
      enabled: true
      prometheusCR:
        enabled: true
      includeStaticScrapeConfigs: false
node:
  metrics:
    cadvisor:
      separate_pipeline: false
```

Results:

- TA discovered the ServiceMonitor — `/jobs` returns
  `serviceMonitor/poc-app/example-app/0`, target assigned to the scraper.
- Scraper polled TA's `/scrape_configs` and `/jobs/.../targets` cleanly,
  no client errors.
- `prometheusremotewrite/observe` exporter shipped metrics with no auth or
  send errors.
- Verified in Observe backend (dataset `42519622`):
  `http_request_duration_seconds_{bucket,count,sum}`, `http_requests_total`,
  `version`, `up`, `scrape_*` series arrived, enriched with
  `k8s.namespace.name=poc-app`, `k8s.deployment.name=example-app`,
  `endpoint=https`, `service.name=example-app`.

## Notes

- TA subchart's ClusterRole already grants
  `monitoring.coreos.com/{servicemonitors,podmonitors}` reads — no extra RBAC.
- TA may log a benign "secrets is forbidden" warning if an SM references a
  secret; no impact on SMs without secret refs.


obs-gh-kaisu (approved):

charts/agent/templates/prometheus-ta-configmap.yaml
```diff
@@ -9,14 +14,28 @@ data:
...
     filter_strategy: relabel-config
     prometheus_cr:
-      enabled: false
+      enabled: {{ $crEnabled }}
+{{- if $crEnabled }}
+      # Label selectors must be present (even empty) — TA treats omitted ones
+      # as "match nothing". Namespace selectors are optional (default = all).
+      service_monitor_selector: {{ toYaml ($ta.prometheusCR.serviceMonitorSelector | default dict) | nindent 8 }}
+      pod_monitor_selector: {{ toYaml ($ta.prometheusCR.podMonitorSelector | default dict) | nindent 8 }}
+      {{- with $ta.prometheusCR.serviceMonitorNamespaceSelector }}
```
obs-gh-mattcotter: The `with` usage here is a bit inconsistent with the above. I think all four values are doing the same thing? If that is the case, please make them consistent.

obs-gh-mattcotter (approved):
One style comment, otherwise lgtm! Thanks for the quick fix!

charts/agent/templates/prometheus-ta-configmap.yaml
sfc-gh-zeningchen: Done. Upstream's defaults differ between label and namespace selectors, which is why I had them
asymmetric. Always emit {} now. Thanks!

PR #507, opened 2026-05-29
fix: gate prometheusCR.enabled on targetAllocator.enabled, simplify TA-CRD example

sfc-gh-zeningchen:
When `targetAllocator.enabled=true`, the chart now forces the merged
receiver path regardless of `cadvisor.separate_pipeline`. This means the
CRD-discovery example no longer needs a `node.metrics.cadvisor` block at
all — keeping the example minimal.

Also adds a render-time gate so `prometheusCR.enabled=true` without
`targetAllocator.enabled=true` fails loudly instead of silently doing
nothing.

Verified that all existing examples (recommended-config, aws-eks-fargate,
prometheus-ta-enabled) render byte-identical before and after this change.


obs-gh-mattcotter (commented):
Thanks! Would you also update the TA CR example as well to remove the CAdvisor block?

sfc-gh-zeningchen:
> Thanks! Would you also update the TA CR example as well to remove the CAdvisor block?

sure, just did!

charts/agent/examples/prometheus-ta-cr-discovery/values.yaml
```diff
@@ -18,5 +18,4 @@ application:
 node:
```
obs-gh-mattcotter: Can this whole block be deleted? I think it's not doing anything now. My opinion is that when the examples are as minimal as possible, they are more valuable to users.

sfc-gh-zeningchen: The issue here is we set the default value of `separate_pipeline` to true to keep the current behavior(separating pod_metrics and cadvisor in 2 separate receivers), and enabling ta requires us to set it to false.  
Perhaps we can make `targetAllocator.enabled` overwrite the value of `separate_pipeline`? so that `separate_pipeline` is only needed when ta is disabled.

obs-gh-mattcotter: Right, thanks for clarifying that. I like that idea! Let's have `targetAllocator.enabled=true` imply `separate_pipeline=false`. Then we still avoid breaking changes and we can simplify this config a bit. Thank you!

sfc-gh-zeningchen: thanks! I've just updated the logic

charts/agent/examples/prometheus-ta-cr-discovery/values.yaml
obs-gh-mattcotter, line 11: One more small comment: I think it would be convenient if `targetAllocator.enabled = true` implied `independentDeployment = true` to cut down on overhead and reduce failure cases.

obs-gh-mattcotter (approved):
One more comment, but I'm okay with that being a follow up if you'd like

charts/agent/examples/prometheus-ta-cr-discovery/values.yaml
sfc-gh-zeningchen: thanks! I'd probably leave it as a follow-up because we used `independentDeployment` as a condition in `Chart.yaml`, it decides whether we deploy the subchart before template is rendered and condition doesn't allow expressions.

PR #523, opened 2026-08-19
fix: Change hardcoded ClusterRoleBinding name

sfc-gh-kvishwanathan:
Since ClusterRoleBinding is cluster-scoped, the previous hard-coded name meant two installs of the chart in different namespaces would both try to create/own a binding named observe-agent-cluster-role-binding, colliding with each other. Suffixing with the namespace makes the name unique per install and keeps it consistent with the ClusterRole name and the binding's own label.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.94.1
+version: 0.94.2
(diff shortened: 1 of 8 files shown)


obs-gh-mattcotter (commented):
lgtm! you will also need to bump the chart patch version, and might have to rebase to a single commit to get CI to pass

sfc-gh-zeningchen (approved):
lgtm! same on bumping the chart version

obs-gh-mattcotter (approved):

PR #524, opened 2026-08-21
fix: make pod-metrics scrape_timeout configurable

sfc-gh-mso:
The `pod-metrics` scrape job sets `scrape_interval` but never `scrape_timeout`, so
every target falls back to Prometheus' 10s default, and no chart value can raise it.
An exporter that needs longer than 10s to answer loses the whole scrape.

That failure is easy to miss. A scrape which runs past its deadline fails the same
way an unreachable target does, so the metrics stop arriving and an operator sees
absence rather than an error.

## What this changes

Adds `application.prometheusScrape.timeout`, rendered as `scrape_timeout` on the
`pod-metrics` job beside the existing `interval`.

**The default is `10s`, the value targets already get today.** The rendered job now
states that default instead of leaving it to Prometheus, so scrape behaviour is
unchanged for every existing consumer. Only someone who sets the field gets
something different.

Prometheus requires `scrape_timeout <= scrape_interval`, which caps how high a
consumer can take this. At the chart's default 60s interval there is room to spare.

## How we ran into it

A `jmx-exporter` sidecar on our own Kafka brokers needs roughly 9-11s to render a
response once a broker carries enough topics. That sits right on the 10s line, so
some scrapes land, some do not, and the broker's JMX metrics go missing with nothing
reporting a failure.

One caveat on those numbers: the scrape durations we can observe are cut off by the
very timeout being measured, so they understate the real distribution rather than
describe it.

## Ancillary edits

The `Chart.yaml` bump (0.94.2 to 0.94.3), the regenerated helm-docs `README.md`, and
the regenerated `make generate-examples` output are all required by this repo's CI
checks rather than changes in their own right.

## Checks run

- `make lint` (`ct lint --all`): all 8 charts pass, including the chart-version bump check
- `make generate-examples`: 4 rendered examples pick up the new line
- `pre-commit` (helm-docs, whitespace, EOF): pass
- This chart carries no `values.schema.json` and no `charts/agent/tests/`, so nothing
  else needed regenerating

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent
 description: Chart to install K8s collection stack based on Observe Agent
 type: application
-version: 0.94.2
+version: 0.94.3
 appVersion: "2.18.0"
 dependencies:
   - name: opentelemetry-collector
--- a/charts/agent/README.md
+++ b/charts/agent/README.md
@@ -1,6 +1,6 @@
 # agent
 
-![Version: 0.94.2](https://img.shields.io/badge/Version-0.94.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.18.0](https://img.shields.io/badge/AppVersion-2.18.0-informational?style=flat-square)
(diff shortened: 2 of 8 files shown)


sfc-gh-zeningchen (approved):
LGTM

PR #525, opened 2026-08-26
feat: allow configurable agent id

sfc-gh-kvishwanathan:


charts/agent/values.yaml
```diff
@@ -522,6 +522,18 @@ cluster-events:
...
+    # Identifies this agent process in Observe. This deployment runs a single
+    # replica, so the deployment name is used because it stays the same when the
+    # pod restarts. Override it if you scale this deployment past one replica,
+    # otherwise every replica reports the same identity.
+    # The deployment name is "<release>-cluster-events". The downward API cannot
+    # read it directly, so the release half comes from the label holding it.
+    - name: DEPLOYMENT_INSTANCE_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.labels['app.kubernetes.io/instance']
```
obs-gh-mattcotter, line 534: Am I correct that this is a workaround for `{{ .Release.Name }}` since we can't use template variables in the `values.yaml`?

charts/agent/values.yaml
```diff
@@ -827,6 +851,18 @@ prometheus-scraper:
           name: agent-credentials
           key: OBSERVE_TOKEN
           optional: true
+    # Identifies this agent process in Observe. This deployment runs a single
+    # replica, so the deployment name is used because it stays the same when the
```
obs-gh-mattcotter, line 855: This isn't always true. We have support for sharding this horizontally with the Target Allocator.  I like this naming when it is a single instance, but when it's sharded, I think we might just need to use pod id.

charts/agent/values.yaml
```diff
@@ -1305,6 +1359,12 @@ forwarder:
           name: agent-credentials
           key: TRACE_TOKEN
           optional: true
+    # Identifies this agent process in Observe. One agent runs per node, and the
```
obs-gh-mattcotter, line 1362: This isn't always true either 😬 We have support for running the forwarder as a deployment.

charts/agent/values.yaml
```diff
@@ -658,6 +670,18 @@ cluster-metrics:
...
+    # pod restarts. Override it if you scale this deployment past one replica,
+    # otherwise every replica reports the same identity.
+    # The deployment name is "<release>-cluster-metrics". The downward API cannot
+    # read it directly, so the release half comes from the label holding it.
+    - name: DEPLOYMENT_INSTANCE_NAME
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.labels['app.kubernetes.io/instance']
+    - name: OBSERVE_AGENT_INSTANCE_ID
+      value: "$(DEPLOYMENT_INSTANCE_NAME)-cluster-metrics"
```
obs-gh-abhishekrao, line 684: Wouldn't this approach lead to conflicting IDs when multiple clusters are being monitored?

charts/agent/values.yaml
sfc-gh-kvishwanathan: ya thats right

PR #526, opened 2026-08-31
feat(agent): add hash processors for external-write Iceberg tables

sfc-gh-vramakrishnan:
Observe source tables carry hidden hash columns that the ingest transformer normally computes: `_observe_hash126_11identifiers` on a k8sEntity table and `_observe_hash126_6labels` on a prometheus table. The query optimizer substitutes `group by <object>` with the stored hash rather than recomputing it, so a writer that appends to such a table directly -- bypassing ingest -- must supply the hashes itself, and the collector is the only place that sees the data first.

## Processor definitions

Adds two processors in `charts/agent/templates/_config-processors.tpl`:

- `transform/identifiers_hash` — hashes `attributes["observe_transform"]["identifiers"]` for K8sEntity rows. Must run last in the pipeline (after `observek8sattributes` + `transform/object` populate the map).
- `transform/metric_hashes` — hashes the union of datapoint attributes and resource attributes for Prometheus rows. Must be the last datapoint-attribute-setting processor before the exporter, and must include resource attributes because `resource_to_telemetry_conversion` promotes them to labels after this processor runs.

Both use `XXH128` truncated to 31 hex chars (124 bits) because the underlying storage column is `decimal(38,0)` (max ~1.0e38); a full 128-bit digest would overflow ~2/3 of the time. The consumer converts hex to decimal at write time.

The hashes need not be bit-identical to Observe's own algorithm — the optimizer trusts whatever is stored — but the algorithm must be deterministic and the dataset must never mix external and internal writes.

## Wiring

Adds `agent.config.global.externalWriteHashes.enabled` (default `false`). When enabled:

- **`transform/identifiers_hash`** appended to the `logs/objects` pipeline in `_cluster-events-config.tpl`.
- **`transform/metric_hashes`** appended to every pipeline that exports to `prometheusremotewrite/observe`:
  - `_cluster-metrics-config.tpl` (metrics pipeline)
  - `_config-pipelines.tpl` (k8s_metrics / pod_metrics / cadvisor)
  - `_node-logs-metrics-config.tpl` (hostmetrics, kubeletstats)
  - `_gateway-config.tpl` (metrics/observe-forward, gated on `outputFormat == prometheus`)
  - `_forwarder-config.tpl` (metrics/observe-forward, gated on `outputFormat == prometheus`)
  - `_monitor-config.tpl` (metrics pipeline)
  - `_prometheus-scraper-config.tpl` (gated on `prometheusScrape.enabled`)
  - `_fargate-sidecar-config.tpl` (metrics/kubeletstats)

Default `false` means rendered output is unchanged on the standard ingest path; users on external-write Iceberg flip the flag once and both hashes are wired everywhere they need to be.

## Example + render-diff coverage

New example `charts/agent/examples/external-write-hashes/` turns the flag on. Its rendered output is checked in, so any future PR that breaks the wiring will surface as a diff in `make generate-examples`.

Note: does not emit `_observe_hash_metric` (the metric-name hash). That one has to be bit-identical to a literal the query compiler inlines at compile time, and the exporter normalizes metric names after every processor runs, so OTTL only ever sees the pre-normalization name. Compute it in the writer, over the metric string actually being written.

Refs: OB-62148

### DIFF
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -509,5 +509,8 @@ transform/metric_hashes:
   metric_statements:
     - context: datapoint
       statements:
-        - set(attributes["observe_labels_hash"], Substring(XXH128(Concat([ToKeyValueString(resource.attributes, "=", ",", true), ToKeyValueString(attributes, "=", ",", true)], "|")), 0, 31))
+        # resource attrs override datapoint attrs due to ordering (matches prometheusremotewrite resource_to_telemetry_conversion)
+        - merge_maps(cache, attributes, "upsert")
+        - merge_maps(cache, resource.attributes, "upsert")
+        - set(attributes["observe_labels_hash"], Substring(XXH128(ToKeyValueString(cache, "=", ",", true)), 0, 31))
 {{- end -}}
--- a/charts/agent/templates/_config-processors.tpl
+++ b/charts/agent/templates/_config-processors.tpl
@@ -469,3 +469,92 @@ Parameters (passed via dict):
       action: update
       new_name: traces.span.metrics.duration.{{ $suffix }}
 {{- end -}}
+
+{{- /*
+Hash processors for EXTERNAL-WRITE Iceberg source tables.
+
+Observe's source tables carry hidden hash columns that the ingest transformer
+normally computes: _observe_hash126_11identifiers on a k8sEntity table and
+_observe_hash126_6labels on a prometheus table. The query optimizer SUBSTITUTES
+`group by <object>` with the stored hash rather than recomputing it, so the
+column must be populated for that optimization to be correct.
+
+A writer that appends to such a table directly -- bypassing ingest -- therefore
+has to supply those hashes itself, and the collector is the only place that sees
+the data before it is written. Neither processor is needed on the normal ingest
+path, where the transformer fills the columns.
+
+Neither is wired into any pipeline by this chart. Add the one you need to the
+relevant pipeline's `processors` list through an agent.config.<workload>
+override, and note the ordering constraints on each below.
+
+The hash need not match Observe's own algorithm; the optimizer trusts whatever
+is stored. It must only be a deterministic function of the same input, and the
+dataset must never mix external and internal writes -- two rows with equal
+identifiers would otherwise carry different hashes and silently fail to group.
+
+Values are 31 hex digits (124 bits) rather than the full 32. The column is
+decimal(38,0), max ~1.0e38, while a full XXH128 reaches ~3.4e38, so roughly
+two thirds of digests would overflow it. The consumer converts hex to decimal.
+*/ -}}
+
+{{- /*
+_observe_hash126_11identifiers, for a k8sEntity source table.
+
+MUST be placed LAST in the pipeline. observek8sattributes and transform/object
+are what populate observe_transform.identifiers, so hashing any earlier digests
+an empty or partial map.
+
+ToKeyValueString's 4th argument sorts the keys, which is what makes this
+deterministic: identifiers is a map and map order is not stable across records
+or agent versions. It serializes keys AND values, so two entities of the same
+kind hash differently -- hashing sorted keys alone would collapse every Pod onto
+one hash.
+
+Emitted as a sibling attribute, deliberately NOT inside
+observe_transform.identifiers: writing it there would mutate the very map being
+hashed, and would also leak the hash into the identifiers column.
+*/ -}}
+{{- define "config.processors.transform.observe_identifiers_hash" -}}
+transform/identifiers_hash:
+  error_mode: ignore
+  log_statements:
+    - context: log
+      statements:
+        - set(attributes["observe_identifiers_hash"], Substring(XXH128(ToKeyValueString(attributes["observe_transform"]["identifiers"], "=", ",", true)), 0, 31))
+{{- end -}}
+
+{{- /*
+_observe_hash126_6labels, for a prometheus source table.
+
+Ordering is load-bearing twice over:
+
+  1. This must be the ONLY datapoint attribute the pipeline sets after it. The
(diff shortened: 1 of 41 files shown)


charts/agent/templates/_config-processors.tpl
```diff
@@ -469,3 +469,45 @@ Parameters (passed via dict):
...
+and the exporter converts metric names to prometheus format, after every processor
+runs. That hash should be calculated in the ingest pipeline itself.
+*/ -}}
+{{- define "config.processors.transform.observe_metric_hashes" -}}
+transform/metric_hashes:
+  error_mode: ignore
+  metric_statements:
+    - context: datapoint
+      statements:
+        - set(attributes["observe_labels_hash"], Substring(XXH128(Concat([ToKeyValueString(resource.attributes, "=", ",", true), ToKeyValueString(attributes, "=", ",", true)], "|")), 0, 31))
```
obs-gh-mattcotter: Should these attribute maps be merged first? Having "service.name=test" in resource attributes vs attributes both look the same in prometheus format (because it gets merged to `labels`). Looks like this hash generates different values. [example](https://ottl.run/#eyJ2ZXJzaW9uIjoidjAuMTU5LjAiLCJleGVjdXRvciI6InRyYW5zZm9ybV9wcm9jZXNzb3IiLCJwYXlsb2FkIjoie1wicmVzb3VyY2VNZXRyaWNzXCI6W3tcInJlc291cmNlXCI6e1wiYXR0cmlidXRlc1wiOlt7XCJrZXlcIjpcInNlcnZpY2UubmFtZVwiLFwidmFsdWVcIjp7XCJzdHJpbmdWYWx1ZVwiOlwidGVzdFwifX1dfSxcInNjb3BlTWV0cmljc1wiOlt7XCJzY29wZVwiOntcIm5hbWVcIjpcIm15LmxpYnJhcnlcIixcInZlcnNpb25cIjpcIjEuMC4wXCJ9LFwibWV0cmljc1wiOlt7XCJuYW1lXCI6XCJteS5jb3VudGVyXCIsXCJ1bml0XCI6XCIxXCIsXCJkZXNjcmlwdGlvblwiOlwiSSBhbSBhIENvdW50ZXJcIixcInN1bVwiOntcImFnZ3JlZ2F0aW9uVGVtcG9yYWxpdHlcIjoxLFwiaXNNb25vdG9uaWNcIjp0cnVlLFwiZGF0YVBvaW50c1wiOlt7XCJhc0RvdWJsZVwiOjIsXCJzdGFydFRpbWVVbml4TmFub1wiOlwiMTU0NDcxMjY2MDMwMDAwMDAwMFwiLFwidGltZVVuaXhOYW5vXCI6XCIxNTQ0NzEyNjYwMzAwMDAwMDAwXCIsXCJhdHRyaWJ1dGVzXCI6W119XX19XX1dfSx7XCJyZXNvdXJjZVwiOnt9LFwic2NvcGVNZXRyaWNzXCI6W3tcInNjb3BlXCI6e1wibmFtZVwiOlwibXkubGlicmFyeVwiLFwidmVyc2lvblwiOlwiMS4wLjBcIn0sXCJtZXRyaWNzXCI6W3tcIm5hbWVcIjpcIm15LmNvdW50ZXJcIixcInVuaXRcIjpcIjFcIixcImRlc2NyaXB0aW9uXCI6XCJJIGFtIGEgQ291bnRlclwiLFwic3VtXCI6e1wiYWdncmVnYXRpb25UZW1wb3JhbGl0eVwiOjEsXCJpc01vbm90b25pY1wiOnRydWUsXCJkYXRhUG9pbnRzXCI6W3tcImFzRG91YmxlXCI6MixcInN0YXJ0VGltZVVuaXhOYW5vXCI6XCIxNTQ0NzEyNjYwMzAwMDAwMDAwXCIsXCJ0aW1lVW5peE5hbm9cIjpcIjE1NDQ3MTI2NjAzMDAwMDAwMDBcIixcImF0dHJpYnV0ZXNcIjpbe1wia2V5XCI6XCJzZXJ2aWNlLm5hbWVcIixcInZhbHVlXCI6e1wic3RyaW5nVmFsdWVcIjpcInRlc3RcIn19XX1dfX1dfV19XX0iLCJjb25maWciOiJ0cmFuc2Zvcm0vbWV0cmljX2hhc2hlczpcbiAgZXJyb3JfbW9kZTogaWdub3JlXG4gIG1ldHJpY19zdGF0ZW1lbnRzOlxuICAgIC0gY29udGV4dDogZGF0YXBvaW50XG4gICAgICBzdGF0ZW1lbnRzOlxuICAgICAgICAtIHNldChhdHRyaWJ1dGVzW1wib2JzZXJ2ZV9sYWJlbHNfaGFzaFwiXSwgU3Vic3RyaW5nKFhYSDEyOChDb25jYXQoW1RvS2V5VmFsdWVTdHJpbmcocmVzb3VyY2UuYXR0cmlidXRlcywgXCI9XCIsIFwiLFwiLCB0cnVlKSwgVG9LZXlWYWx1ZVN0cmluZyhhdHRyaWJ1dGVzLCBcIj1cIiwgXCIsXCIsIHRydWUpXSwgXCJ8XCIpKSwgMCwgMzEpKSJ9)

obs-gh-virjramakrishnan: Right you are! Let me do that. 

It would result in a very slight behavior change: in the current scheme, the conversion to prom will cause resource attributes to override datapoint attributes (under `resource_to_telemetry_conversion: true`). If two datapoints had a different datapoint attribute, and they were overridden to the same value, then my current code would have them as two different hashes, but the proposed change would provide the hash as the same. I think this is desired as this is how the hash would work were it to be in the ingest pipeline, since it receives the map already merged. Besides, the customer wouldn't be able to see the difference because the map is merged in Observe anyway.

obs-gh-mattcotter (approved):

PR #528, opened 2026-09-05
fix(agent): keep the ingest token out of the observe-agent ConfigMap

craig-rueda:
## Problem

`charts/agent/templates/_observe-agent.tpl` renders

```yaml
token: {{ .Values.observe.token.value }}
```

into the `observe-agent` **ConfigMap**, so the Observe ingest token is stored in plain text. Tooling that masks Secrets does not mask ConfigMaps — the token is plainly visible in the ArgoCD UI, in `kubectl get cm observe-agent -o yaml`, and to anyone with `get configmap` in the namespace. The chart already puts the same token in the `agent-credentials` Secret and exposes it to pods as an env var, which is the right mechanism; the ConfigMap copy is an additional, unmasked place the credential lives.

## The ConfigMap value is redundant

observe-agent calls `viper.AutomaticEnv()` with `SetEnvPrefix("OBSERVE")` deliberately commented out ([`internal/root/root.go`](https://github.com/observeinc/observe-agent/blob/v2.18.0/internal/root/root.go)), so the config key `token` resolves from the **`TOKEN`** env var — and viper checks env *before* the config file. Every collector in this chart already gets `TOKEN` from the `agent-credentials` Secret via `extraEnvs` (7 occurrences in `values.yaml`, one per collector), so the Secret value is the one that wins today and the ConfigMap copy is never actually used.

Confirmed against the real binary rather than by reading the code — config file says one thing, env says another:

```console
$ cat observe-agent.yaml
observe_url: https://123456789.collect.observeinc.com
token: TOKEN_FROM_CONFIGMAP

$ docker run --rm -v "$PWD:/conf:ro" -e TOKEN=TOKEN_FROM_SECRET_ENV \
    --entrypoint /observe-agent observeinc/observe-agent:2.18.0 \
    config --observe-config=/conf/observe-agent.yaml
# ======== computed agent config
...
token: TOKEN_FROM_SECRET_ENV
```

And with the line removed entirely, the agent still resolves the token and validates:

```console
$ helm template t charts/agent -f charts/agent/ci/test-values.yaml --set observe.token.create=true \
    | yq -e 'select(.kind == "ConfigMap" and .metadata.name == "observe-agent").data.relay' > observe-agent.yaml
$ docker run --rm -v "$PWD:/conf:ro" -e TOKEN=fake-ds:fake-token \
    --entrypoint /observe-agent observeinc/observe-agent:2.18.0 \
    config validate --observe-config=/conf/observe-agent.yaml
✅ configuration is valid
```

Same result on 2.12.0 and 2.18.0.

## The change

Rather than dropping the line outright, it is emitted only when the chart is *not* managing the Secret and the value would otherwise be lost, so there is no behaviour change for anyone who has opted out of the chart-managed Secret:

| `observe.token.create` | `observe.token.value` | ConfigMap `token` | Secret `OBSERVE_TOKEN` |
| --- | --- | --- | --- |
| `true` | set | **absent** (was plaintext) | present |
| `false` | set | present — unchanged | absent |
| `false` | empty | absent (was an empty `token:`) | absent |

The default path in every `examples/` and `ci/` values file is `create: true`, so those all stop leaking the token; the `create: false` + explicit value combination keeps working exactly as before.

Happy to drop the fallback branch entirely if you would rather make this unconditional in a minor/major bump — that combination is already documented as "assumed that a secret named `agent-credentials` with key `OBSERVE_TOKEN` already exists", so arguably it should never have been supported.

## Also included

- `charts/agent/Chart.yaml` version bump 0.94.3 → 0.94.4
- Note in `values.yaml` explaining when the token stays out of the ConfigMap
- Regenerated `examples/*/rendered/observe-agent-configmap.yaml` goldens and `README.md` (helm-docs); the only README delta is the version badge
- `helm lint` clean against all three `ci/*.yaml` values files

## Not included

`templates/nodeless-otelcollector.yaml` still interpolates `Bearer {{ .Values.observe.token.value }}` into the Fargate `OpenTelemetryCollector` CR. Same class of exposure, but fixing it means reworking the sidecar config to read `${env:TOKEN}` and wiring the Secret into the CR, so I left it out to keep this focused. Glad to follow up if you want it.


obs-gh-justindaines (approved):
Looks good. I manually tested on a local `kind` cluster and AWS Fargate.

